PR #1, opened 2017-04-18
An initial overview of threads

binji:
For now this is just copied from https://github.com/WebAssembly/design/pull/1019, since I haven't had a chance to work on the proposal much. I'll address some of the feedback there soon.

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -311,15 +311,15 @@ new Memory(memoryDescriptor)
 
 However, the `memoryDescriptor` now will check for a `shared` property:
 
-If [HasProperty][]("shared"), then let `shared` be
-[ToBoolean][]([Get][](`memoryDescriptor`, "shared")). Otherwise, let `shared`
+If [HasProperty][](`"shared"`), then let `shared` be
+[ToBoolean][]([Get][](`memoryDescriptor`, `"shared"`)). Otherwise, let `shared`
 be `false`.
 
 Let `memory` be the result of calling [Memory.create][] given arguments
-`initial`, `maximum`, and `shared`. Note that initial and maximum are specified
-in units of WebAssembly pages (64KiB).
+`initial`, `maximum`, and `shared`. Note that `initial` and `maximum` are
+specified in units of WebAssembly pages (64KiB).
 
-Return the result of [CreateMemoryObject][](`memory`).
+Return the result of [`CreateMemoryObject`](#creatememoryobject)(`memory`).
 
 ### `CreateMemoryObject`
 
@@ -335,8 +335,8 @@ If `m` is shared, any attempts to [detach][] `buffer` shall throw a
 [TypeError][]. Otherwise, any attempts to [detach][] `buffer` _other_ than the
 detachment performed by `m.grow` shall throw a [TypeError][].
 
-Return a new `WebAssembly.Memory` instance with [[Memory]] set to `m` and
-[[BufferObject]] set to `buffer`.
+Return a new `WebAssembly.Memory` instance with `[[Memory]]` set to `m` and
+`[[BufferObject]]` set to `buffer`.
 
 ### `WebAssembly.Memory.prototype.grow`
 
@@ -557,6 +557,7 @@ instr ::= ...
 [Memory.create]: https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.ml#L47
 [Memory.memory]: https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli#L1
 [Memory.grow]: https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.ml#L60
+[SharedArrayBuffer]: https://tc39.github.io/ecma262/#sec-arraybuffer-objects
 [SharedArrayBuffer]: https://tc39.github.io/ecma262/#sec-sharedarraybuffer-objects
 [detach]: http://tc39.github.io/ecma262/#sec-detacharraybuffer
 [TypeError]: https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror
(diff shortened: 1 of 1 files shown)


binji:
I've updated the overview to not reference things from the design repo. I've also tried to add some basic spec changes, though perhaps those would be better to just do directly in the rst files.

Still haven't gotten around to the JavaScript-stuff.

AndrewScheidecker:
The alignment immediate on these operators is effectively ignored, right? If so, I think it should be removed (or specification of unnatural alignment should be made invalid).

jfbastien:
> The alignment immediate on these operators is effectively ignored, right? If so, I think it should be removed (or specification of unnatural alignment should be made invalid).

Elsewhere we discussed trapping if not at least naturally-aligned. Given that, I tend to agree that we could just drop alignment entirely.

binji:
Makes sense, but I believe we were going to use the alignment u32 as the place to add a flag/index for accessing memories other than 0, right? So perhaps we should just keep the alignment but require it to be natural.

proposals/threads/Overview.md
```diff
@@ -1 +1,480 @@
...
+share  ::= notshared | shared
+```
+
+Its [encoding][limits encoding] is as follows:
+
+```
+limits ::= 0x00 n:u32          => {min n, max e, notshared}
+           0x01 n:u32 m:u32    => {min n, max m, notshared}
+           0x11 n:u32 m:u32    => {min n, max m, shared}
+```
```
AndrewScheidecker: Is the shared-unbounded case deliberately omitted? (i.e. `0x10 n:u32          => {min n, max e, shared}`)

And what is the text syntax?

binji: Yes, it is deliberately omitted. I can't remember what issue we discussed it, but it simplifies things if we can just assume the VM will reserve address space for the maximum size. Though I suppose if grow_memory is just best effort anyway, maybe it doesn't really make a difference.

Haven't really thought of a text syntax yet. I guess it could just be:
```
(memory <min> <max> shared)
```

proposals/threads/Overview.md
```diff
@@ -1 +1,480 @@
...
+    change.
+
+## Wait and Wake operators
+
+The wake and wait operators are optimizations over busy-waiting for a value to
+change. It is a validation error to use these operators on non-shared linear
+memory. The operators have sequentially consistent ordering.
+
+Both wake and wait operators trap if the effective address of either operator
+is misaligned or out-of-bounds.
```
AndrewScheidecker: "Misaligned" for `wake` means 4 byte aligned, right? This should be spelled out, or perhaps `wake` shouldn't care about alignment.

proposals/threads/Overview.md
jfbastien: We're worried about having to suspend all threads if we need to physically move the memory. If max is specified we know we'll never need to move the memory because the virtual memory was pre-allocated, and only page protections or bounds need to be changed which doesn't require thread suspension.

It's easy to add later, so I'm not too worried. It would be good to document, and point at the experience from NaCl. IIRC the issues around suspension were documented in its codebase.

proposals/threads/Overview.md
```diff
@@ -1 +1,480 @@
...
+### Resizing
+
+If linear memory is marked as shared, the maximum memory size must be
+specified.
+
+`grow_memory` and `current_memory` have [sequentially consistent][] ordering.
+
+`grow_memory` of a shared linear memory is allowed, but may fail if the new
+size is greater than the specified maximum size, or if reserving the additional
+memory fails. All agents in the executing agent's cluster will have access to
```
sunfishcode: This wording makes it a little unclear that `grow_memory` always fails if the new size is greater than the specified maximum size, which is what I assume is intended here. Possible wording suggestion:

`grow_memory` of a shared linear memory is allowed. Like non-shared memory, it fails if the new size is greater than the memory's maximum size, and it may also fail if reserving the additional memory fails.

proposals/threads/Overview.md
binji: I can't seem to find the place in the NaCl source that talks about the difficulty in thread suspension. But I'll add a comment about it generically, and we can add further references later.

proposals/threads/Overview.md
binji: Good point, I've explicitly mentioned wake requires 4-byte alignment.

proposals/threads/Overview.md
binji: Thanks, I've used this wording.

binji:
OK, first pass of the JS API changes. PTAL

proposals/threads/Overview.md
```diff
@@ -297,6 +297,69 @@ returns the number of waiters that were woken as an `i32`.
...
+
+If `m` is shared, let `buffer` be a new [SharedArrayBuffer][]. If `m` is not
+shared, let `buffer` be a new [ArrayBuffer][]. In either case, `buffer` will
+have an internal slot [\[\[ArrayBufferData\]\]][] which aliases `m` and an
+internal slot [\[\[ArrayBufferByteLength\]\]][] which is set to the byte length
+of `m`.
+
+If `m` is shared, any attempts to [detach][] `buffer` shall throw a
+[TypeError][]. Otherwise, any attempts to [detach][] `buffer` _other_ than the
+detachment performed by `m.grow` shall throw a [TypeError][].
```
jfbastien: To confirm: shared memory grow detatches the SAB?

binji: No, I guess this should be clearer. The "Otherwise" is meant to refer to when `m` is not shared; a SAB is never detached.

proposals/threads/Overview.md
```diff
@@ -1 +1,571 @@
...
+
+The WebAssembly.Memory constructor has the same signature:
+
+```
+new Memory(memoryDescriptor)
+```
+
+However, the `memoryDescriptor` now will check for a `shared` property:
+
+If [`HasProperty`][](`"shared"`), then let `shared` be
```
rossberg-chromium: The usual ES conventions is to not use `HasProperty` in such situations (e.g. because it is an observable extra call with proxies), but handle `undefined` as a result of `Get` appropriately. Fortunately, `undefined`, already converts to `false`, so you can just drop the check without substitute.

rossberg-chromium:
Wait, when has the opcode prefix changed to F0? As long as the only motivation for that remains [being silly](https://github.com/WebAssembly/design/pull/1019#issuecomment-291421051), it still is not worth the price of unnecessary opcode space fragmentation IMO.

binji:
r.e. opcode prefix: I don't really care either way, but both you and @jfbastien  have argued for different codes. It doesn't seem like there's a strong technical argument either way, and we don't have a policy for opcode layout. Perhaps we should? This sort of thing is going to keep coming up.

binji:
Aside from the opcode prefix, are there any other issues that I should address before landing this overview?

binji:
OK, landing w/ lgtm.

PR #14, opened 2017-05-23
Write up alternate encoding using i8/i16 value types

binji:
As described here: https://github.com/WebAssembly/design/issues/1049

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -749,6 +749,7 @@ The compare exchange operators are:
 ## Opcode Count Comparisons
 
 | Proposal # | Description | # New Non-Atomic Ops | # New Atomic Ops | New Ops / RMW |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


stoklund:
This is also relevant for the SIMD `extract_lane` instructions.

binji:
Agreed, I mention that [here](https://github.com/WebAssembly/threads/pull/14/files#diff-e12c2aa9d5fae657e6d0c78964ef2f95R578). :-)

stoklund:
Ah, apologies. I missed that. :-)

You can get much of the same effect without introducing new value types by defining the equivalent of LLVM's SelectionDAG `zext_inreg` operators. It's basically your `i32.extend_*/i*` operator, but the input value is a `i32` where the high bits are ignored. This also maps directly to `utxb` and `uxtw` instructions on ARM.

binji:
@stoklund good point, I'll mention that.

binji:
Landing, assuming this accurately reflects the alternate encodings. If not, please open a new issue.

PR #22, opened 2017-05-29
Update WebAssembly.Global definition

binji:


lukewagner:
While you were clarifying globals, I had a few requests, reading just now:
* It seems like `Global.prototype.value` should be an accessor property with null setter and getter that returns the value of the global.  This is the way to express properties (another example being `Memory.prototype.buffer` whose values are magically changed (in a way other than the user setting the property)
* For global imports, I think we should keep mutable and const globals completely separate: if you have a mutable global import, you *must* pass a `WebAsembly.Global` (else `LinkError`) and if you have a const global import, you *must* pass a Number (else `LinkError`).  I think anything else more coercive could cover up linking errors.
* For mutable global imports, there is no need to throw on `i64` because there are no JS numbers involved.  Instead, the `Global.prototype.value` getter can throw if the gotten value is an `i64`.

rossberg-chromium:
Change LGTM, though I agree with the additional points @lukewagner is making. I actually think it would be fine for immutable imports to allow both plain numbers as a shorthand, or immutable Globals, but I don't feel strongly either way. However, if Global are to only represent mutable globals then I'd suggest removing the @@toPrimitive hook, since that's an implicit conversion as well, and even over a stateful entity.

lukewagner:
What I hadn't considered that @binji pointed out today was that, with the `toPrimitive` hook and auto-boxing of global imported primitive values, we could probably pull of the technically-breaking change of having *all* imports/exports take/produce `WebAssembly.Global` objects.  This seems increasingly attractive to me.

binji:
Landing w/ LGTM and assumption that @lukewagner is OK with it too. :-)

I'll continue to iterate in further PRs.

PR #27, opened 2017-06-06
Update Overview.md to match decisions from meeting

binji:
* Remove float atomics
* ~~Remove i64.wait~~
* Use 0xfe as prefix byte
* Add sign-extending conversion operators
* Remove sign-extending atomic-operators (encoding proposal 3)

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -255,6 +255,7 @@ is woken, the wait operator returns 0 ("ok"). If the timeout expires before
 another agent wakes this one, this operator returns 2 ("timed-out").
 
   * `i32.wait`: load i32 value, compare to expected (as `i32`), and wait for wake at same address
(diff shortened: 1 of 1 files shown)


jfbastien:
Concensus on i64.wait was neutral, slightly positive. You should keep it.

binji:
@jfbastien I'm fine keeping it, but I wonder how to interpret the CG decision in this case. Does "neutral" mean it's up to the discretion of the champion, or does it mean that there's no interest in the CG so the champion shouldn't bother? I think I know your answer, but it does seem like something we should be clear about :-)

jfbastien:
Neutral with a mild preference means you can remove or keep, but people slightly wanted to keep. It also means next meeting you could be asked to remove it! It's the worst kind of outcome: you didn't get much guidance.

PR #31, opened 2017-06-19
Add clearer examples to Globals.md for rationale

binji:


proposals/threads/Globals.md
```diff
@@ -2,6 +2,316 @@
...
+  (import "env" "memory" (memory 1))
+  ...
+)
+```
+
+Then the modules can be instantiated as follows:
+
+```
+WebAssembly.instantiate(m1Bytes, {}).then(
+    ({_, instance}) => {
```
domenic: Here and elsewhere, it should just be `({instance})`; no `_,` necessary. (That tries to destructure a non-existant property `_` from the result, so `_` is always `undefined`.)

proposals/threads/Globals.md
```diff
@@ -2,6 +2,316 @@
...
+    ;; SP = SP + 4
+    (set_global $sp (i32.add (get_global $sp (i32.const 4))))
+  )
+)
+```
+
+With the modules instantiated as follows:
+
+```
+let agentSp = 0x100;
```
domenic: This example is interesting because it doesn't show any usage of `WebAssembly.Global`. Is that a mistake? Or do we need another example to show the value-add of that?

proposals/threads/Globals.md
binji: Done.

proposals/threads/Globals.md
binji: Yeah, I should add an example for that too.

binji: Done.

proposals/threads/Globals.md
```diff
@@ -2,6 +2,341 @@
...
+
+```
+let agentIndex = 0;
+let spAddrs = [0x0, 0x4];
+let imports = {env: {spAddr: spAddrs[agentIndex]}};
+WebAssembly.instantiate(m1Bytes, imports).then(
+    ({instance}) => {
+        let imports = {env: {
+          memory: instance.exports.memory,
+          spAddr: spAddrs[agentIndex],
```
hammacher: Is there a "++agentIndex" missing somewhere?

proposals/threads/Globals.md
```diff
@@ -2,6 +2,341 @@
...
+* All function calls that load/store SP must spill SP at call sites and
+  function entrypoints
+* Additional thread-local values must also be spilled/loaded in the same way to
+  have the same benefit
+* The SP is still in linear memory, so it can easily be trashed by another
+  agent
+
+### Solution 3: Modify Function Signature to Pass SP as Parameter
+
+Rather than storing spilling the SP to linear memory, the SP value can be
```
hammacher: Typo: "storing spilling"

proposals/threads/Globals.md
```diff
@@ -2,6 +2,341 @@
...
+passed as a parameter. Because we can't tell whether an imported function will
+use the SP, we must modify all imported and exported functions.
+
+The SP will ultimately be saved in a mutable global, but will be loaded from
```
hammacher: Why even store it in a mutable global? Why not just use the local / parameter?

binji: It's just an optimization. You could always use the parameter, but the idea here was to only modify exported functions' signatures.

proposals/threads/Globals.md
binji: A different `agentIndex` would be in another worker... so this example is a bit confusing. I'll clarify.

Issue #24, opened 2017-05-29
WebAssembly.Memory JS API tweaks


lukewagner:
Two requests for `WebAssembly.Memory`:
* `Memory` constructor should describe what is thrown if `shared` is set but `maximum` is not given.
* The `grow` method only updates the current JS object suggesting that other `Memory` objects in other actors that alias the same underlying memory will hold their old SABs with now-out-of-date lengths.  Rather than trying to talk about actively updating all of them at once (which has interesting atomicity questions), what if instead the `Memory.prototype.buffer` getter does a lazy check: if the current SAB's byteLength doesn't match the current memory's length, a new SAB object is created and the slot is updated?

PR #32, opened 2017-06-21
Tweak WebAssembly.Memory definition

binji:
See comments in #24:

* Throw an error when constructing shared memory without maximum.
* Lazily create SharedArrayBuffer after grow.

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -251,6 +251,9 @@ However, the `memoryDescriptor` now will check for a `shared` property:
 Let `shared` be [`ToBoolean`][]([`Get`][](`memoryDescriptor`, `"shared"`)).
 Otherwise, let `shared` be `false`.
 
+If `shared` is `true`, and [`HasProperty`][](`"maximum"`) is `false`, then a
+[`TypeError`][] is thrown.
+
 Let `memory` be the result of calling [`Memory.create`][] given arguments
 `initial`, `maximum`, and `shared`. Note that `initial` and `maximum` are
 specified in units of WebAssembly pages (64KiB).
@@ -290,14 +293,31 @@ Let `d` be [`ToNonWrappingUint32`][](`delta`).
 
 Let `ret` be the current size of memory in pages (before resizing).
 
-Perform [`Memory.grow`][] with delta `d`. On failure, a [`RangeError`][] is thrown.
-
-Assign to `M.[[BufferObject]]` a new [`SharedArrayBuffer`][] whose
-[\[\[ArrayBufferData\]\]][] aliases `M.[[Memory]]` and whose
-[\[\[ArrayBufferByteLength\]\]][] is set to the new byte length of
-`M.[[Memory]]`.
+Perform [`Memory.grow`][] with delta `d`. On failure, a [`RangeError`][] is
+thrown.
 
 Return `ret` as a Number value.
+
+### `WebAssembly.Memory.prototype.buffer`
+
+This is an accessor property whose [[Set]] is Undefined and whose [[Get]]
+accessor function performs the following steps:
+
+1. If `this` is not a `WebAssembly.Memory`, a [`TypeError`][] is thrown.
+1. Otherwise:
+  1. If `m` is not shared, then return `M.[[BufferObject]]`.
+  1. Otherwise:
+    1. Let `newByteLength` be the byte length of `M.[[Memory]]`.
+    1. Let `oldByteLength` be
+       `M.[[BufferObject]].`[\[\[ArrayBufferByteLength\]\]][].
+    1. If `newByteLength` is equal to `oldByteLength`, then return
+       `M.[[BufferObject]]`.
+    1. Otherwise:
+      1. Let `buffer` be a new [`SharedArrayBuffer`][] whose
+         [\[\[ArrayBufferData\]\]][] aliases `M.[[Memory]]` and whose
+         [\[\[ArrayBufferByteLength\]\]][] is set to `newByteLength`.
+      1. Set `M.[[BufferObject]]` to `buffer`.
+      1. Return `buffer`.
 
 ## [Spec Changes][spec]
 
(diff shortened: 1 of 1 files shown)


lars-t-hansen:
So about creating a new SAB object... SAB objects are mutable, it's possible to add properties to them.  When the length is increased, and a new SAB object is created, the new object will not have any properties that had been added to the old buffer object.  This is fine.  But that only happens if the length is *increased*, not if `grow_memory` is called with argument 0.  I'm not yet convinced this is what we want, because that `0` can be the result of some (complicated) computation.  We are now in a situation where most of the time, but occasionally not, `grow_memory` will appear to clear out expandos on the buffer.  This seems brittle to me.  It would be more user friendly to return a new SAB object iff `grow_memory` has been observed on the memory.

(Implementation-wise a 64-bit serial number, stored in the memory and incremented on `grow_memory`, and copied into the SAB object and compared to what's in the memory when the buffer is requested, will be adequate; it will never wrap around on current systems.)

lukewagner:
Good point!  An alternative approach would be to have a [[Generation]] counter on the shared wasm memory (bumped atomically on every `grow_memory`) and a [[Generation]] slot on the per-agent `Memory` JS object and new SABs are issued (and the per-agent [[Generation]] slot updated to match the wasm memory [[Generation]]) when the two don't match.

lukewagner:
Yes, good point and the counter makes sense.  I think it would be stored on the (per-agent) `WebAssembly.Memory` JS object, not the SAB, though.

binji:
I guess I don't see why it's better to know that `grow_memory` causes your cached SAB to be invalidated vs. actually growing the memory. You could argue that `grow_memory(0)` should be a no-op, so why would it invalidate your cached SAB?

If I understand your concern, it is that someone will write this:
```
let instance = ... // some WebAssembly instance.
let memory = instance.exports.memory;  // Shared memory.
memory.buffer.foo = 3;
instance.myGrowMemory(0);  // Calls grow_memory(0).
memory.buffer.foo;  // Still 3.
```
Then will try the same thing but call `myGrowMemory(1)` and be surprised that foo is gone?

I think realistically they'll actually call:
```
instance.myBlackBoxFunction();  // Does this call grow_memory at all?
memory.buffer.foo;  // Whoops, it does -- this returns undefined.
```
So they can just as easily hit this bug whether it is `grow_memory(0)` or `grow_memory(n)`, right?

The workaround seems somewhat invasive too, since we'll need to store a counter on the internal memory object that is only used by the embedder.

Isn't the combination of `grow_memory(0)` and storing properties on the SAB sufficiently rare enough to argue against somewhat invasive workarounds to prevent this developer mistake?

lukewagner:
I agree that this probably won't trip up people in practice; if people are going to be tripped up it's when the value of `.buffer` *is* changed to a new SAB, not when it stays the same.  Mostly, I just agreed with Lars that it seems oddly noncontiguous.

But perhaps one way to rationalize the `byteLength`-equality behavior is that the cached SAB is only updated when it *needs* to be and if the current SAB still aliases the whole of memory, it doesn't need to be.

Another interesting point is that, because `grow_memory` can happen in other agents, it's quite racy which SAB you'll get on any given invocation of the `.buffer` getter.  It seems like *any* use of expandos would be hazardous and I was thinking maybe the SAB returned by `.buffer` should be frozen.  Identity is still observable (via `===`) and expandos are still possible (via `WeakMap`), so this would just be a deterrent.

binji:
> It seems like any use of expandos would be hazardous and I was thinking maybe the SAB returned by .buffer should be frozen.

I like this idea. Is it common to automatically freeze an object in JavaScript APIs?

lukewagner:
Not that I know.  Another alternative that is non-racy in object identity (just racy in `byteLength` and contents :) is for `.buffer` to *always* return a new SAB.  This would ensure noone could mistakenly assume it didn't change.  It'd also maintain that all data-raciness was confined to SAB reads/writes.

binji:
> Another alternative that is non-racy in object identity (just racy in byteLength and contents :) is for .buffer to always return a new SAB.

Yeah, I thought about that too. It seems nasty to require all those allocations, but I guess this is JS so we don't really care much. :-)

Another thing I just realized: we're requiring the developer to reload or recreate the buffers in every imported function that accesses memory:
```
let memory = ...;
function myImport(addr) {
  // If addr is a pointer into recently acquired memory, we need to make sure we get a new SAB. 
  let u8a = new Uint8Array(memory.buffer);
}
```
Caching the buffer won't do the right thing by default. Specifically, it means that the standard emscripten `HEAP8`, `HEAP32`, etc. should be reloaded at every function import.

lars-t-hansen:
Regarding whether `0` should mean one thing or the other thing: for non-shared memory, `Memory.p.grow(0)` does detach the old buffer object, ie, `0` is treated uniformly.  Weird for it not to be treated so here.

Anyhow, setting that aside, either freezing the SAB or returning a new object every time seems fine.  Freezing the SAB will trip some people up and will lead to some rioting and accusations of working against the spirit of JS, though it's the view objects that will matter in practice, so this seems like a very minor concern.  Old SAB objects can't become detached on grow, so old SAB objects can be cached in performance-sensitive code, mitigating the cost of returning a new object every time.  IMO we should go with new-object-every-time.

domenic:
> Is it common to automatically freeze an object in JavaScript APIs?

It's not that common, because this situation comes up pretty rarely. But it's done in a few instances, e.g. template string parts and various web APIs where they want to return a property but are returning a usually-mutable object. So yeah, this seems like a reasonable idea to me.

> Another alternative that is non-racy in object identity (just racy in byteLength and contents :) is for .buffer to always return a new SAB.

Please do not do this. https://w3ctag.github.io/design-principles/#attributes-like-data

binji:
> Regarding whether 0 should mean one thing or the other thing: for non-shared memory, Memory.p.grow(0) does detach the old buffer object, ie, 0 is treated uniformly. Weird for it not to be treated so here.

Oh, huh. Yeah, I see what you mean.

> Please do not do this. https://w3ctag.github.io/design-principles/#attributes-like-data

This is pretty clear. So I guess freezing is the best bet. I'll modify the text to mention this, unless there are other concerns.

binji:
PTAL, added wording about freezing the buffer.

binji:
Merging with lgtm and general agreement.

Issue #36, opened 2017-06-26
Float timeout for wait


rossberg-chromium:
The `wait` operator takes its timeout operand as a float. It is not clear to me how this could be implemented correctly for large values. Unless I'm missing something, the implementation would need a timer that has the value range of a large float combined with the resolution of a small float _over that entire value range_. How would this be handled?

Btw, if a float is used anyway, why not stick to the metric system properly and use seconds as the base unit?

binji:
> The wait operator takes its timeout operand as a float. It is not clear to me how this could be implemented correctly for large values.

AIUI, the typical assumption of timeouts are that they provide a guarantee that it won't be less than the specified value, so it is OK for the implementation to round up to the granularity they provide. My quick calculations show that we have nanosecond precision for timeouts of up to 142 years with a 64-bit float given we use milliseconds as the base unit. Not sure any OS can actually provide timeouts at that granularity anyway.

> Btw, if a float is used anyway, why not stick to the metric system properly and use seconds as the base unit?

I chose milliseconds to match ES `Atomics.wait`. If we use seconds, then the above calculation is reduced to ~50 days at nanosecond granularity. Still plenty, I think.

Also: the primary reason to use float was to provide a way to specify an infinite timeout without having a special value. That said, we already treat negative values as special for `wait` so perhaps it's more consistent that way.

lars-t-hansen:
And to round that out, `Atomics.wait` uses a millisecond base units to match existing APIs that use that, notably performance.now() but there are probably others; I had the impression this representation is pretty standard.

ISTR very early versions of the SAB proposal used an integer number of nanoseconds or microseconds, but with only 32-bit integers in ES that's not very appealing.  In ES, "infinite" timeout could be handled by passing `undefined` or other OOB value, even if it's clunky.

IMO nothing stops us from having a different API for Wasm than for ES, and an int64 number of nanoseconds might be just fine, but I think changing the base unit from milliseconds to seconds while still keeping the argument a floating point number is just inviting trouble.

domenic:
https://w3ctag.github.io/design-principles/#milliseconds

rossberg-chromium:
I'm still a bit confused how that is supposed to be handled. AFAICT, even
OS-level time(r)s are integers everywhere, e.g. C and POSIX have `time_t`,
which is in seconds. POSIX timers use `timespec`, which is a pair of a
`time_t` and a long nanoseconds value. If I were to map a float timeout to
[...]

PR #43, opened 2017-06-30
Change timeout from f64 ms -> i64 ns

binji:
See issue #36.

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -204,15 +204,13 @@ was suspended (e.g. via the embedder or a wait operator).
 ### Wait
 
 The wait operator take three operands: an address operand, an expected value,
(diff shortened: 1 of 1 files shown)


jfbastien:
I think we need to justify why we'd diverge from https://w3ctag.github.io/design-principles/#milliseconds that @domenic points out. This is weirdly inconsistent with the JS side (both ns and using integer instead of float). I don't really understand the argument for this change: perceived implementation complexity is already there for JS, so the change saves engines zero. The discrepancy does, however, add complexity to users because JS and wasm would now be different.

binji:
OK, please voice this opinion on the issue.

binji:
We agreed in the CG meeting to use i64 nanoseconds. Any other issues with this PR?

titzer:
lgtm

PR #45, opened 2017-06-30
Merge spec

binji:


binji:
Oops, forgot that this squashes by default. Will reset and merge manually.

Issue #16, opened 2017-05-24
Unordered initialisation from data segments


rossberg-chromium:
The overview currently says about initialisation from data segments:

> When the imported linear memory is shared, the writes are non-atomic and are not ordered.

What's the rationale for unordered? Is that necessary or desirable? It would make data segments significantly more non-deterministic than everything else proposed. Unlike with a general memcpy operator the memory regions cannot overlap here, so isn't it easy enough to guarantee an ordering?

lars-t-hansen:
Don't know about the motivation for that language, but if the writes are not atomic, they may appear unordered to another observer anyway.  If another observer wants to observe them predictably some sync after initialization will make the order of writes moot - the writes will all be visible to the observer at that point.

rossberg-chromium:
I see, makes sense.

However, is there actually a way in this proposal to synchronise after such initialisations? Since they are not performed by user code, how could sychronisation be expressed in user code? Would it be sufficient to synchronise in the start function under the given memory model?

lars-t-hansen:
If the backing buffer is distributed to workers after initialization has run then synchronization should be implicit in that operation (per the ES spec).  But if the backing buffer is distributed before initialization, it should be enough for the initializing thread (which presumably must start first) to write a word atomically at the start of its operation and for the other threads to atomically read that word when they start and observe the correct value there.

That said, this all probably depends on who performs initialization in a multi-threaded setting, and when that happens, and how it's controlled.  I confess I'm not completely clear on that, yet.

jfbastien:
Consider a case where:

1. Create a WebAssembly.Memory with the shared attribute
2. Instantiate a module with this memory
3. Much later, instantiate a different module with the same memory

When we do 3. there can be data segments which get initialized, and the instance from 2. can notice that initialization.

The intent of the proposal is that all data segment initialization is non-atomic, and that once that's complete the initialized writes are guaranteed visible (they happen-before 3. invokes the start function). This means that data segment initialization needs not be performed with atomic writes and need only use a single barrier (after all segment initialization is performed, or before an exception is thrown because OOB segment).

Maybe the proposed wording doesn't actually do this?

It also sounds like @rossberg-chromium isn't sure these are the right semantics. Would you instead rather initialize all data segments atomically? At which granularity? "Byte or more" sounds simplest. Maybe this can be another poll for the meeting.

@binji can you clarify the wording as above, and if @rossberg-chromium agrees also add a poll?

binji:
[...]

PR #46, opened 2017-07-05
Update text on Data Segment initialization

binji:
See issue #16.

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -59,7 +59,7 @@ non-shared linear memory:
 * From low to high bytes
 * At byte granularity (which can be coalesced)
 * As non-atomics
-* An entire module's data segment initialization then synchronizes with other
+* An entire module's data section initialization then synchronizes with other
   operations (effectively, followed by a barrier)
   
 The intention is to allow the implementor to "memcpy" the initializer data into
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -48,12 +48,12 @@ agent's cluster will have access to the additional linear memory.
 When a module has an imported linear memory, its data segments are copied into
 the linear memory when the module is instantiated.
 
-When linear memory is not shared, the ordering of data segment initialization is
-not observable. When linear memory is shared, it is possible for another module
-(or in the web embedding, for JavaScript code) to read from the linear memory as
-it is being initialized.
+When linear memory is shared, it is possible for another module (or in the web 
+embedding, for JavaScript code) to read from the linear memory as it is being
+initialized.
 
-The data segments are therefore initialized as follows:
+The data segments are initialized as follows, whether they apply to shared or
+non-shared linear memory:
 
 * Data segments are initialized in their definition order
 * From low to high bits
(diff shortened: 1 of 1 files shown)


proposals/threads/Overview.md
```diff
@@ -48,8 +48,48 @@ agent's cluster will have access to the additional linear memory.
 When a module has an imported linear memory, its data segments are copied into
 the linear memory when the module is instantiated.
 
-When the imported linear memory is shared, the writes are non-atomic and
-are not ordered.
+When linear memory is not shared, the ordering of data segment initialization is
+not observable. When linear memory is shared, it is possible for another module
```
jfbastien: That's not true: you can observe an OOB segment in the corresponding memory (or rather, observe all the segments which were initialized before it OOB'd). What's true is that you can't observe order within a segment.

I think what you specify below could just apply to non-shared as well.

binji: Perhaps I misunderstanding what you mean, but the JS.md doc is pretty explicit that there [cannot be a failure partway through intialization](https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-objects):

> Segments may overlap and, if they do, the final value is the last value written in order. Note: there should be no errors possible that would cause this operation to fail partway through.

jfbastien: Oh right, I'd forgotten that we'd done that change (and I'd fixed our impl).

binji: That said, I can specify the ordering for non-shared as well; do you think it is valuable to do so?

jfbastien: Yeah I think it's simpler to just specify it for both. Even if non-observable, the mental model is simpler.

binji: Done.

proposals/threads/Overview.md
```diff
@@ -48,8 +48,48 @@ agent's cluster will have access to the additional linear memory.
...
-are not ordered.
+When linear memory is shared, it is possible for another module (or in the web 
+embedding, for JavaScript code) to read from the linear memory as it is being
+initialized.
+
+The data segments are initialized as follows, whether they apply to shared or
+non-shared linear memory:
+
+* Data segments are initialized in their definition order
+* From low to high bits
```
rossberg-chromium: Did you mean bytes?

binji: I admit I just copied this from JF's comment, but yes, I think bytes makes more sense here. :-)

proposals/threads/Overview.md
```diff
@@ -48,8 +48,48 @@ agent's cluster will have access to the additional linear memory.
...
+
+The data segments are initialized as follows, whether they apply to shared or
+non-shared linear memory:
+
+* Data segments are initialized in their definition order
+* From low to high bytes
+* At byte granularity (which can be coalesced)
+* As non-atomics
+* An entire module's data segment initialization then synchronizes with other
+  operations (effectively, followed by a barrier)
```
jfbastien: I think you mean "and entire module's data *section* [...]" ? Since each data section contains multiple segments.

PR #47, opened 2017-07-07
Structured{De,}Serialize of WebAssembly.Memory

binji:


proposals/threads/Overview.md
```diff
@@ -333,6 +333,25 @@ For the web embedding, `wake` is equivalent in behavior to executing the followi
 
 ## [JavaScript API][] changes
 
+### Structured Serialization and Deserialization of `WebAssembly.Memory`
```
rossberg-chromium: Nit: Strictly speaking, these are changes to the Web API, not the JS API.

domenic:
> What does it mean to serialize a shared Memory? Especially as things are being written to it.

HTML is pretty clear on this. The serialization record just contains a pointer to the shared memory. This pointer is then passed across agent boundaries, etc.

There are separate specs, like IndexedDB, which define the process of snapshotting a serialization record into bytes on disk. (Or rather, at this time, I don't believe it's well-defined, but when the spec editors have more bandwidth, they plan to do so.)

binji:
Ah, I see. I was a bit concerned that this proposal didn't mention it anywhere, so I wanted to put something in. I guess I can just wait for that and reference it directly.

lukewagner:
Coincidentally just talking to @lars-t-hansen about this today: if you try to send a non-`shared` `Memory`, could that perhaps just throw instead of making a copy?  B/c of the heap-mapping tricks, this may end up having surprising performance effects.

binji:
Closing in favor of https://github.com/WebAssembly/design/pull/1074.

PR #53, opened 2017-07-27
Fix a missing paren

winksaville:
In section "Initializing Memory Only Once" the WebAssembly.instantiate is missing a paren.


binji:
Thanks!

PR #54, opened 2017-07-27
Add spec text for extend operators

binji:
See published version here: https://webassembly.github.io/threads

PR #55, opened 2017-07-28
Atomic memory accesses

binji:
Doesn't include execution yet.

### DIFF
--- a/document/core/binary/instructions.rst
+++ b/document/core/binary/instructions.rst
@@ -286,6 +286,7 @@ Each variant of :ref:`atomic memory instruction <syntax-instr-atomic-memory>` is
      \hex{FE}~\hex{4E}~~m{:}\Bmemarg &\Rightarrow& \I64.\ATOMICRMW\K{32\_u}.\ATOMICCMPXCHG~m \\
    \end{array}
 
+
 .. index:: numeric instruction
    pair: binary format; instruction
 .. _binary-instr-numeric:
--- a/document/core/binary/instructions.rst
+++ b/document/core/binary/instructions.rst
@@ -160,6 +160,132 @@ Each variant of :ref:`memory instruction <syntax-instr-memory>` is encoded with 
    In future versions of WebAssembly, the additional zero bytes occurring in the encoding of the |CURRENTMEMORY| and |GROWMEMORY| instructions may be used to index additional memories.
 
 
+.. index:: atomic memory instruction
+   pair: binary format; instruction
+.. _binary-instr-atomic-memory:
+
+Atomic Memory Instructions
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Each variant of :ref:`atomic memory instruction <syntax-instr-atomic-memory>` is encoded with a different byte code. Loads, stores and RMW instructions are followed by the encoding of their |memarg| immediate.
+
+.. _binary-atomic-load:
+.. _binary-atomic-loadn:
+.. _binary-atomic-store:
+.. _binary-atomic-storen:
+.. _binary-atomic-rmw:
+.. _binary-atomic-rmwn:
+
+.. math::
+   \begin{array}{llclll}
+   \production{instruction} & \Binstr &::=& \dots \\ &&|&
+     \hex{FE}~\hex{10}~~m{:}\Bmemarg &\Rightarrow& \I32.\ATOMICLOAD~m \\ &&|&
+     \hex{FE}~\hex{11}~~m{:}\Bmemarg &\Rightarrow& \I64.\ATOMICLOAD~m \\ &&|&
+     \hex{FE}~\hex{12}~~m{:}\Bmemarg &\Rightarrow& \I32.\ATOMICLOAD\K{8\_u}~m \\ &&|&
+     \hex{FE}~\hex{13}~~m{:}\Bmemarg &\Rightarrow& \I32.\ATOMICLOAD\K{16\_u}~m \\ &&|&
+     \hex{FE}~\hex{14}~~m{:}\Bmemarg &\Rightarrow& \I64.\ATOMICLOAD\K{8\_u}~m \\ &&|&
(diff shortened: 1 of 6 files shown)


binji:
No preview for this one yet, wanted to keep this separate from #54 for now. But it would be nice to get an idea if this looks good so far. One thing I was a bit concerned about was the fact that `atomicop` includes `cmpxchg`, but I separate those out in the validation text.

document/core/binary/instructions.rst
```diff
@@ -160,6 +160,132 @@ Each variant of :ref:`memory instruction <syntax-instr-memory>` is encoded with
...
+   \phantom{\production{instruction}} & \phantom{\Binstr} &\phantom{::=}& \phantom{\dots} && \phantom{thisshouldbeenough} \\[-2ex] &&|&
+     \hex{FE}~\hex{48}~~m{:}\Bmemarg &\Rightarrow& \I32.\ATOMICRMW.\ATOMICCMPXCHG~m \\ &&|&
+     \hex{FE}~\hex{49}~~m{:}\Bmemarg &\Rightarrow& \I64.\ATOMICRMW.\ATOMICCMPXCHG~m \\ &&|&
+     \hex{FE}~\hex{4A}~~m{:}\Bmemarg &\Rightarrow& \I32.\ATOMICRMW\K{8\_u}.\ATOMICCMPXCHG~m \\ &&|&
+     \hex{FE}~\hex{4B}~~m{:}\Bmemarg &\Rightarrow& \I32.\ATOMICRMW\K{16\_u}.\ATOMICCMPXCHG~m \\ &&|&
+     \hex{FE}~\hex{4C}~~m{:}\Bmemarg &\Rightarrow& \I64.\ATOMICRMW\K{8\_u}.\ATOMICCMPXCHG~m \\ &&|&
+     \hex{FE}~\hex{4D}~~m{:}\Bmemarg &\Rightarrow& \I64.\ATOMICRMW\K{16\_u}.\ATOMICCMPXCHG~m \\ &&|&
+     \hex{FE}~\hex{4E}~~m{:}\Bmemarg &\Rightarrow& \I64.\ATOMICRMW\K{32\_u}.\ATOMICCMPXCHG~m \\
+   \end{array}
+
```
rossberg-chromium: Nit: use 2 emtpy lines before new subsection.

document/core/syntax/instructions.rst
```diff
@@ -265,6 +265,58 @@ Both instructions operate in units of :ref:`page size <page-size>`.
...
+
+The |ATOMICRMW| instructions are `read-modify-write <https://en.wikipedia.org/wiki/Read-modify-write>`_
+instructions. They each have an :ref:`atomicop <syntax-instr-atomicop>`, which
+specifies how memory will be modified. Each instruction returns the value read
+from memory before modification. The |ATOMICXCHG| :ref:`atomicop <syntax-instr-atomicop>`
+doesn't use the read value, but instead stores its argument unmodified. The
+|ATOMICCMPXCHG| :ref:`atomicop <syntax-instr-atomicop>` is similar, but only
+performs this action conditionally, if the read value is equal to a provided
+comparison argument. All other :ref:`atomicops <syntax-instr-atomicop>` have
+behavior of the :ref:`ibinop <syntax-ibinop>` of the same name.
```
rossberg-chromium: Nit: 2 lines

document/core/text/instructions.rst
```diff
@@ -214,6 +214,92 @@ Lexically, an |Toffset| or |Talign| phrase is considered a single :ref:`keyword
...
+.. _text-atomic-loadn:
+.. _text-atomic-store:
+.. _text-atomic-storen:
+.. _text-atomic-rmw:
+.. _text-atomic-rmwn:
+
+The offset immediate to atomic memory instructions is optional, and defaults to
+:math:`\T{0}`.
+
+.. math::
```
rossberg-chromium: It might be necessary to split this math up in the same manner as for numeric instructions, in order to allow suitable page reaks in the PDF.

document/core/text/instructions.rst
```diff
@@ -214,6 +214,92 @@ Lexically, an |Toffset| or |Talign| phrase is considered a single :ref:`keyword
...
+   \begin{array}{llclll}
+   \production{instruction} & \Tplaininstr_I &::=& \dots \\ &&|&
+     \text{i32.atomic{.}load}~~m{:}\Tmemarg_4 &\Rightarrow& \I32.\ATOMICLOAD~m \\ &&|&
```
rossberg-chromium: Nit: the braces around the dot shouldn't be needed inside a \text arg.

document/core/valid/instructions.rst
```diff
@@ -424,6 +424,178 @@ Memory Instructions
      C \vdash \GROWMEMORY : [\I32] \to [\I32]
    }
```
rossberg-chromium: Nit: add extra line

document/core/text/instructions.rst
binji: Yeah, this was strange -- I found that without this Mathjax would produce extra space after `.`, probably because it assumes that it is the end of a sentence. I suppose it doesn't do this for `32.` because it thinks its a number?

document/core/text/instructions.rst
binji: Ah, I see. I was wondering why that was done elsewhere. I find the misalignment of the columns to be really ugly, and I assumed that the phantom stuff was meant to try to fix that, but it didn't seem to be working for me.

rossberg-chromium: Hm, the instruction names are much longer in this case, so maybe the "thisisenough" text isn't actually enough here?

document/core/text/instructions.rst
rossberg-chromium: Ah. Yeah, I've seen that ugliness, too, and I don't understand why Mathjax does it. Seems like a bug to me, it doesn't happen with proper Latex (and the \mathsf environment isn't representing text). Good to know that the braces fix it, I might want to do that elsewhere as well then

PR #58, opened 2017-08-09
Add `share` flag to `memtype`

binji:
The binary encoding is a little awkward since we need to store the `share` bit inside of limits. I originally wrote this up by adding the `share` bit to the limits struct, but it makes everything quite a bit more complicated.

Rendered version [here](https://webassembly.github.io/threads):

### DIFF
--- a/document/core/binary/types.rst
+++ b/document/core/binary/types.rst
@@ -78,8 +78,10 @@ Limits
 .. math::
    \begin{array}{llclll}
    \production{limits} & \Blimits &::=&
-     \hex{00}~~n{:}\Bu32 &\Rightarrow& \{ \LMIN~n, \LMAX~\epsilon \} \\ &&|&
-     \hex{01}~~n{:}\Bu32~~m{:}\Bu32 &\Rightarrow& \{ \LMIN~n, \LMAX~m \} \\
+     \hex{00}~~n{:}\Bu32 &\Rightarrow& \{ \LMIN~n, \LMAX~\epsilon \}, 0 \\ &&|&
+     \hex{01}~~n{:}\Bu32~~m{:}\Bu32 &\Rightarrow& \{ \LMIN~n, \LMAX~m \}, 0  \\ &&|&
+     \hex{02}~~n{:}\Bu32 &\Rightarrow& \{ \LMIN~n, \LMAX~\epsilon \}, 1 \\ &&|&
+     \hex{03}~~n{:}\Bu32~~m{:}\Bu32 &\Rightarrow& \{ \LMIN~n, \LMAX~m \}, 1 \\
    \end{array}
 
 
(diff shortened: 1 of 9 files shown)


document/core/valid/types.rst
```diff
@@ -88,7 +88,7 @@ Table Types
 Memory Types
 ~~~~~~~~~~~~
 
-:math:`\limits`
+:math:`\limits~\share`
 ...............
```
rossberg-chromium: Nit: probably need to adjust underline?

document/core/valid/modules.rst
```diff
@@ -396,7 +396,6 @@ Imports :math:`\import` and import descriptions :math:`\importdesc` are classifi
 
 
 :math:`\IDMEM~\memtype`
-.......................
```
rossberg-chromium: Why remove the underline?

document/core/util/math.def
```diff
@@ -136,6 +136,8 @@
 .. |ETMEM| mathdef:: \xref{syntax/types}{syntax-externtype}{\K{mem}}
 .. |ETGLOBAL| mathdef:: \xref{syntax/types}{syntax-externtype}{\K{global}}
 
+.. |MSHARED| mathdef:: \xref{syntax/types}{syntax-memtype}{\K{shared}}
+.. |MNOTSHARED| mathdef:: \xref{syntax/types}{syntax-memtype}{\K{notshared}}
```
rossberg-chromium: Maybe `unshared` reads a bit nicer.

document/core/binary/types.rst
```diff
@@ -90,14 +90,21 @@ Limits
...
 ~~~~~~~~~~~~
 
-:ref:`Memory types <syntax-memtype>` are encoded with their :ref:`limits <binary-limits>`.
+:ref:`Memory types <syntax-memtype>` are encoded with their :ref:`limits <binary-limits>` that includes an extra value to specify whether the the memory is shared.
 
 .. math::
    \begin{array}{llclll@{\qquad\qquad}l}
    \production{memory type} & \Bmemtype &::=&
-     \X{lim}{:}\Blimits &\Rightarrow& \X{lim} \\
+     \hex{00}~~n{:}\Bu32 &\Rightarrow& \{ \LMIN~n, \LMAX~\epsilon \}~\MNOTSHARED \\ &&|&
```
rossberg-chromium: Instead of this "code duplication", can the binary grammar for limits perhaps return an additional flags byte representing the other bits of the opcode? Then you just need to enforce that to be zero in the table case. (Potentially, table types might grow some flags as well in the future, so that it could even be useful to make it reusable.)

binji: I like this idea, but I'm not sure quite how to represent this in the doc. It seems that most of the "functions" (such as `limits` above) return a specific syntax type, not an arbitrary struct or pair of values. Also, how can I extract multiple values at the callsite (such as in `memtype` or `tabletype`)?

document/core/valid/types.rst
binji: Done.

document/core/valid/modules.rst
binji: Whoops, fixed.

document/core/util/math.def
binji: Done.

document/core/binary/types.rst
rossberg-chromium: There are a couple of other rules that synthesise a pair, e.g. [type uses](https://webassembly.github.io/spec/text/modules.html#type-uses) in the text format. At the use site I just pattern-match the pair, see e.g. [import descriptions](https://webassembly.github.io/spec/text/modules.html#text-importdesc).

binji: Ah, that makes sense. Thanks!

binji:
OK, fixed the code duplication, PTAL

PR #59, opened 2017-08-11
Fix inconsistency with wait timeout value in example

bnjbvr:
The spec says that wait's third argument is an i64, but the example shows a f64, probably following a previous version of the spec.


binji:
Thanks!

PR #60, opened 2017-08-15
Implement extend instructions in interpreter

binji:


rossberg-chromium:
This looks good, with one nit: as defined, the AST includes the operator i32.extend32, which isn't valid. That's fine, and is the case for some other operators already. But even if the parser does not generate it, there should be a check in the validator to rule it out, like for load (see `check_memop` in valid.ml).

binji:
Thanks, see fix in #64.

PR #64, opened 2017-08-28
Check for invalid extend operation

binji:


interpreter/valid/valid.ml
```diff
@@ -138,6 +138,11 @@ let type_cvtop at = function
 
 (* Expressions *)
 
+let check_unop unop at =
+  match unop with
+  | Values.I32 I32Op.Extend32S -> error at "invalid unary op"
```
rossberg-chromium: Nit: s/op/operator/

Issue #62, opened 2017-08-25
Can we avoid data segment fill at every instantiation?


juj:
Reading https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#initializing-memory-only-once, with

```
The data segments are always copied into linear memory, even if the same module 
is instantiated again in another agent. One way to ensure that linear memory is only 
initialized once is to place all data segments in a separate module that is only 
instantiated once, then share the linear memory with other modules.
```

if I understand this correctly, this will prevent Emscripten and other compilers from being able to place data and code sections in the same `.wasm` file, but must always use two separate files for them, similar to how `.mem` files are being used in asm.js.

It sounds like it would be easy to add a `Module.instantiate(..., noDataSegment: true);` type of machinery to avoid this? Given that there's an explicit paragraph about this in the document, I presume this approach has been considered already before, and was turned down? I'm curious, what was the conversation about it back then?

lars-t-hansen:
It's good of you to bring that up again.  The previous discussion started here: https://github.com/WebAssembly/threads/issues/16#issuecomment-304600033.  There was IMO no consensus around the matter, but the matter was closed (by @binji) for reasons that are not clear to me.

juj:
The drive to have one `.wasm` file that contains both the code and the data sections is commonly driven by requests from the community, i.e. some Emscripten users prefer to have as few files in the output as possible, to the point that even embedding the .wasm into a .html file or a .js file has been asked. 

Technically we can definitely make do with two files, one `.wasm` with a `.mem` file for data section, which I think we do already support for wasm as well, though I'm sure people will ask if they can optimize away the `.mem` file when it suddenly becomes two files again. It can be a bit easier to manage caching if there is only one file instead of two, and one has less risk of loading incompatible versions of the two.

Another technical note is that since we'll need to keep the Module in memory throughout the lifetime of the wasm application in order to be able to accommodate `pthread_create` of new threads, it will be more efficient to separate the two files, so the memory initializer/data segment can be dropped at startup after it has been applied, and only the wasm module with the code still kept around. (I'm sure that will not stop people from asking to have only one file instead of two)

Even with the this optimization aspect noted, if it's all the same, I think it would be useful to have a flag for this, especially considering if it is simple to implement in practice. CC @kripken

jfbastien:
I suggest designing the details of this feature and presenting this design to the CG.

binji:
[...]

PR #65, opened 2017-08-29
Add proposal for skipping data segment initialization

binji:
This adds a new proposal doc, with two potential solutions as discussed in #62.

### DIFF
--- a/proposals/threads/DataSegmentInitialization.md
+++ b/proposals/threads/DataSegmentInitialization.md
@@ -1,0 +1,135 @@
+# Data segment initialization flag proposal
+
+This page describes a proposal for providing a flag to skip data segment
+initialization when instantiating a module.
+
+## Rationale
+
+Under the current threading proposal, to share a module between multiple
+agents, the module must be instantiated multiple times: once per agent.
+Instantiation initializes linear memory with the contents in the module's data
+segments. If the memory is shared between multiple agents, it will be
+initialized multiple times, potentially overwriting stores that occurred after
+the previous initializations.
+
+For example:
+
+```webassembly
+// The module.
+(module
+  (memory (export "memory") 1)
+
+  ;; Some value used as a counter.
+  (data (i32.const 1) "\0")
+
+  ;; Add one to the counter.
+  (func (export "addOne")
+    (i32.store8
+      (i32.const 0)
+      (i32.add
+        (i32.load8_u (i32.const 0))
+        (i32.const 1)))
+  )
+)
+```
+
+```javascript
+// main.js
+let moduleBytes = ...;
+
+WebAssembly.instantiate(moduleBytes).then(
+  ({instance}) => {
+    // Increment our counter.
+    instance.exports.addOne();
+
+    // Spawn a new Worker.
+    new Worker('worker.js');
+  });
+
+// worker.js
+let moduleBytes = ...;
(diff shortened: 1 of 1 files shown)


proposals/threads/DataSegmentInitialization.md
```diff
@@ -0,0 +1,135 @@
...
+initialized from the contents of the module's data segments.
+
+## Solution 2: Provide an additional initializer expression per data segment
+
+The [binary format for the data section](https://webassembly.github.io/spec/binary/modules.html#data-section)
+currently has a collection of segments, each of which has a memory index, an
+initializer expression for its offset, and its raw data.
+
+Since WebAssembly currently does not allow for multiple memories, the memory
+index must be zero. We can repurpose this field as a flags field and use it to
```
jfbastien: Repurpose as in make it impossible to specify multiple memories? Or something else?

winksaville: If it means not allowing multiple memories "forever", that doesn't seem wise. IMHO solution 1 looks to be a better approach.

binji: Sorry, I should make this more clear. The intention is that it becomes a flags field, but one of the bits can be used to specify that there exists a non-zero memory index, which would be a separate field. I'll add that to the proposal.

proposals/threads/DataSegmentInitialization.md
```diff
@@ -0,0 +1,135 @@
+# Data segment initialization flag proposal
+
+This page describes a proposal for providing a flag to skip data segment
```
lukewagner: Although it's not technically an issue yet b/c tables can't be shared, it would be good to keep elem and data segments symmetric and make the same change to both.

proposals/threads/DataSegmentInitialization.md
```diff
@@ -0,0 +1,135 @@
...
+  ({instance}) => {
+    // Increment our counter.
+    instance.exports.addOne();
+
+    // Spawn a new Worker.
+    new Worker('worker.js');
+  });
+
+// worker.js
+let moduleBytes = ...;
```
lukewagner: Mostly a nit, but if the example becomes long-lived, it'd be nice if main.js would postMessage the module to the worker which the worker would receive via onmessage.

binji:
Addressed comments, PTAL

lukewagner:
Beautiful, thanks!  lgtm as a proposal

jfbastien:
Yeah as a proposal this is good. I like 2 better.

proposals/threads/DataSegmentInitialization.md
```diff
@@ -0,0 +1,177 @@
...
+
+For example:
+
+```webassembly
+// The module.
+(module
+  (memory (export "memory") 1)
+
+  ;; Some value used as a counter.
+  (data (i32.const 1) "\0")
```
pepyakin: Is there should be `i32.const 0`? IIUC, `addOne` expects counter to be at 0.

binji: Oops, thanks!

pepyakin: @binji fixed https://github.com/WebAssembly/threads/pull/66 )

PR #66, opened 2017-08-30
Fix example in ConditionalSegmentInitialization.md

pepyakin:
It looks like `addOne` func expects counter to be at address 0. 
But data declaration puts counter at address 1. 

Also, changed comment from `//` to `;;` to be in line with other comments in WebAssembly example.

### DIFF
--- a/proposals/threads/ConditionalSegmentInitialization.md
+++ b/proposals/threads/ConditionalSegmentInitialization.md
@@ -19,7 +19,7 @@ the previous initializations.
 For example:
 
 ```webassembly
-// The module.
+;; The module.
 (module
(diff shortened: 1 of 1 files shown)


binji:
Sorry, beat you to it! :-) https://github.com/WebAssembly/threads/commit/0aea0ad8d747e4e50ba9b3ceda57ccc65cfa7519

pepyakin:
Yeah, you are fast indeed! What about changing comment `//` to be `;;`? (but feel free to close!)

binji:
Oh, you're right, this is in a WebAssembly code block. Thanks!

PR #67, opened 2017-09-01
Implement atomic instructions in the interpreter

binji:
This is a WIP for now; it would be useful to have some direction on the best ways to implement this.


binji:
@rossberg-chromium Would you mind giving me some direction w/ this change? I'm avoiding the issue of handling multiple threads for now -- not sure that will actually work just yet.

rossberg-chromium:
@binji, sorry, I haven't forgotten, but didn't get to it yet, since I'm still travelling.

interpreter/binary/decode.ml
```diff
@@ -200,6 +200,74 @@ let memop s =
   let offset = vu32 s in
   Int32.to_int align, offset
 
+let atomic_prefix s =
```
rossberg-chromium: Nit: perhaps `atomic_instr`, since the function names all refer to what's being decoded, not what preceded it.

binji:
Oops, I think I had a question about how best to implement the RMW instructions in `memory.ml`, but I can't quite remember what issue I was having. :-)

Guess I'll just try writing it again and see...

lars-t-hansen:
General comment: I assume it'll be possible to distinguish traps-because-unaligned and traps-because-out-of-bounds, from the error message in the thrown host exception if nothing else.  It'll be useful if we spec this so that if an access is both unaligned and oob, it always traps with the unaligned error, ie, in the spec sense an alignment check happens first.  This will benefit implementations that perform explicit bounds checking and should not be a big problem on implementations that rely on hardware alignment checking, say.

rossberg-chromium:
@lars-t-hansen, the core Wasm semantics does not distinguish different traps. The JS API spec does not prescribe error message payload either -- just like EcmaScript doesn't. JS generally does not support (portable) programmatic distinction of individual error causes. We would need to bend over backwards to make this one a special case.

lars-t-hansen:
@rossberg-chromium, if traps are completely undistinguishable then that's a fine solution.  That said, there's a potential real-life interop issue if implementations allow them to be distinguished anyway.  Maybe what I'm asking for is for the spec to at least be consistent in the order it performs the checks for all the atomic operations, even if that doesn't necessarily have much teeth.

(Presumably actual CPUs have confronted this issue already, although the x64 reference leaves the order of protection faults and alignment faults unspecified for MOV, from what I can tell.)

jfbastien:
I assume @littledan will distinguish trap -> exception conversion in the JS spec, and ordering of traps will be part of what he specifies?

rossberg-chromium:
@jfbastien, not that I know. Ecma 262 considers the error message carried by an exception purely diagnostic, and never specifies its content. That obviously implies that even ordering the checks in certain ways has no normative effect on the chosen message. The same situation extends to the JS API, so I don’t know how we would go about that, or that it is worthwhile (there are plenty of operations in JS itself that can generate many different errors).

littledan:
Formally, what @rossberg-chromium says is true: If we throw the same exception type, and the message isn't specified, then the specification wouldn't technically say which order they come in.

At least in JS, the design philosophy is that the specific error messages are implementation-specific so that implementations can always improve in the specificity of error messages. But the only ones that the conformance tests can check for is when there's a difference in the broad classes.

I haven't heard of interop issues on the web due to exceptions being grouped/reported differently. I've heard of Node programs which had compat issues because of an error message rewording from one version to the next, but I guess such a program wouldn't make it five minutes on the open web due to variability between browsers. Overall, this might not be the most profitable area to focus on locking down for cross-browser compat.

In my development of the binding so far, I was only thinking about specifying the ordering of exceptions that came up in the JS/Web binding itself, and not thinking about how to differentiate these classes.

However, it's also of course possible to specify an ordering of checks, and the JS spec has tons of observable check orderings. It might take some changes in the Wasm spec to accomplish this, if Wasm only exposes to embedders that a generic "trap"/"error" has occurred.

The [current draft text](https://littledan.github.io/spec/document/JS.html#traps) that explains the mapping is lifted from the JS.md text earlier. It doesn't get any more specific--everything at runtime is just a RuntimeError. Should it be more specific?

rossberg-chromium:
The RuntimeError semantics has long been shipping, so there isn't much leeway for introducing a more refined error hierarchy at this point. (And we made that design choice consciously.)

littledan:
@rossberg-chromium Is there any trace of that earlier discussion that I could read? FWIW I think we could both be backwards compatible and introduce a way to differentiate, like DOMError, but I am not sure if we want to copy that.

rossberg-chromium:
@littledan, not sure, you may find some trace on the design repo, but most of it may have happened during prior in-person discussions.

jfbastien:
> However, it's also of course possible to specify an ordering of checks, and the JS spec has tons of observable check orderings. It might take some changes in the Wasm spec to accomplish this, if Wasm only exposes to embedders that a generic "trap"/"error" has occurred.

That sounds like a good way forward.

interpreter/exec/int.ml
```diff
@@ -173,6 +174,8 @@ struct
     let n = clamp_rotate_count y in
     or_ (Rep.shift_right_logical x n) (Rep.shift_left x (Rep.bitwidth - n))
 
+  let xchg x y = y
```
rossberg: Shouldn't this return x?

interpreter/syntax/ast.ml
```diff
@@ -60,6 +62,15 @@ type 'a memop =
 type loadop = (Memory.mem_size * Memory.extension) memop
 type storeop = Memory.mem_size memop
 
+module RmwOp =
```
rossberg: Is it possible to get rid of this by directly returning a rmwop from the lexer?

binji: Done.

interpreter/exec/int.ml
binji: No, we want to return the value to store, which is the new value, `y`. It does seem a bit weird to have in int.ml, though, so I moved in inline to `rmwop`.

PR #76, opened 2017-10-26
Add `wake` and `ixx.wait` instructions to spec doc

binji:
Syntax, validation, binary and text sections are completed. Execution section is still TODO.

### DIFF
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -292,8 +292,8 @@ Instructions in this group are concerned with accessing :ref:`linear memory <syn
      \ATOMICXCHG \\
    \production{instruction} & \instr &::=&
      \dots ~|~ \\&&&
-     \WAKE~\memarg ~|~ \\&&&
-     \K{i}\X{nn}\K{.}\WAIT~\memarg ~|~ \\&&&
+     \ATOMICWAKE~\memarg ~|~ \\&&&
+     \K{i}\X{nn}\K{.}\ATOMICWAIT~\memarg ~|~ \\&&&
      \K{i}\X{nn}\K{.}\ATOMICLOAD~\memarg ~|~ \\&&&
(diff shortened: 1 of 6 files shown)


document/core/syntax/instructions.rst
```diff
@@ -327,6 +329,13 @@ this action conditionally, if the read value is equal to a provided comparison
 argument. All other :ref:`atomicops <syntax-atomicop>` have behavior of the
 :ref:`ibinop <syntax-ibinop>` of the same name.
 
+The |WAKE| and |WAIT| instructions provide primitive synchronization between
+:ref:`threads <syntax-thread>`. The |WAIT| instructions atomically load a value
+from the calculated *effective address* and compare it to an expected value. If
```
rossberg: Nit: no need to emphasize "effective address" here, since the defining occurrence already was above.

binji:
> Looks good. One design question, though: shouldn't the names of these instructions include the atomic prefix like the others?

I suppose so. My original thought was that the `atomic` was necessary to differentiate from `i32.load` and `i32.store`, and the rmw ops got it for consistency, but `wait` and `wake` didn't really need it. But another thought is that `atomic` should be used for any operation with the atomic binary prefix.

rossberg:
Such visual classification would be particularly useful in the context of "compartmentalising" the language as we've discussed at the July meeting.

jfbastien:
Adding "atomic" is probably worth a quick poll, should be easy to get through.

binji:
Sounds good, I'll move forward with the name change for now (since I think it's probably better), and roll back if the CG disagrees.

jfbastien:
@binji can you add an agenda item for this?

binji:
The poll in the CG meeting slightly favored switching to using the `atomic` prefix, so I'll leave that in the spec.

PR #79, opened 2017-11-15
Add alternate proposal for segment init.

binji:
See the discussion in issue #62 starting [here](https://github.com/WebAssembly/threads/issues/62#issuecomment-342456572).

This solution introduces new instructions to initialize segments programmatically, but only in the start function.

### DIFF
--- a/proposals/threads/ConditionalSegmentInitialization.md
+++ b/proposals/threads/ConditionalSegmentInitialization.md
@@ -175,3 +175,58 @@ elem    ::= 0x01 e_o:expr e_a:expr y*:vec(funcindex) => {table 0, offset e_o, ap
 As with data segments, the `apply` initializer expression is evaluated during
 instantiation, and will be applied only if the expression evaluates to a
 non-zero value.
+
+## Solution 3: New instructions to initialize data and element segments
+
+Similar to solution 2, we repurpose the memory index as a flags field. Unlike
+solution 2, the flags field specifies whether this segment is _inactive_. An
+inactive segment will not be automatically copied into the memory or table on
+instantiation, and must instead be applied manually using two new instructions:
+`init_memory` and `init_table`.
+
+When the least-significant bit of the flags field is `1`, the segment is inactive. The rest of the bits of
+the flags field must be zero.
+
+The data section is encoded as follows:
+
+```
+datasec ::= seg*:section_11(vec(data))    => seg
+data    ::= 0x00 e:expr b*:vec(byte)      => {data 0, offset e, init b*, active true}
+data    ::= 0x01 e:expr b*:vec(byte)      => {data 0, offset e, init b*, active false}
+```
+
+The element section would be encoded similarly.
+
+### `init_memory` and `init_table`
+
+These instructions copy data from a given segment into the memory or table. The
+segment index is provided as an immediate value. Both instructions have no
+operands and return no results. If the segment index immediate is
+out-of-bounds, it is a validation error.
+
+When `init_memory` is executed, its behavior exactly matches the steps
+described in step 11 of
+[instantiation](https://webassembly.github.io/spec/exec/modules.html#instantiation).
+Similarly, `init_table` has behavior matching step 10.
+
+These instructions may only be used in the [start
+function](https://webassembly.github.io/spec/syntax/modules.html#start-function),
+and only when the start function is called automatically during instantiation.
+At any other times, the instructions will trap.
+
+Consider the example given in solution 2; there are two data sections, the
+first is always active and the second is conditionally active if global 0 has a
+non-zero value. This could be implemented as follows:
+
+```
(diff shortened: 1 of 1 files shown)


lukewagner:
Thanks for writing this up!  A few high-level thoughts:
* it seems useful to have the destination offset of `init_table`/`init_memory` be a dynamic operand (rather than being specified as the `init_expr` in the data section)
  * if we do this, then the `init_expr` operand of a data segment only has meaning when the segment is marked "active" and thus we should probably not even have that field present when "inactive"
* I can also imagine it being useful (and no real challenge) to additionally have a second `length` operand to `init_table`/`init_memory` that specifies how much of the data segment to copy

proposals/threads/ConditionalSegmentInitialization.md
```diff
@@ -175,3 +175,58 @@ elem    ::= 0x01 e_o:expr e_a:expr y*:vec(funcindex) => {table 0, offset e_o, ap
...
+segment index is provided as an immediate value. Both instructions have no
+operands and return no results. If the segment index immediate is
+out-of-bounds, it is a validation error.
+
+When `init_memory` is executed, its behavior exactly matches the steps
+described in step 11 of
+[instantiation](https://webassembly.github.io/spec/exec/modules.html#instantiation).
+Similarly, `init_table` has behavior matching step 10.
+
+These instructions may only be used in the [start
```
rossberg: Since finding out whether the code is running during instantiation requires a runtime check anyway it doesn't actually seem necessary to distinguish the start function syntactically -- and in fact, it might want to call helpers, for example.

The simplest semantics would come down to this: an instance has two states: initialising vs initialised. After the start function has returned the instance transitions from former to latter. In the second state the init instructions trap.

rossberg:
Agreed with @lukewagner. I'd go even further and say that these instructions should work like memcopy, i.e., have the following arguments:

* source segment (immediate)
* target table/memory (immediate)
* source offset (i32 arg)
* target offset (i32 arg)
* length (i32 arg)

lars-t-hansen:
Agreed with @lukewagner and @rossberg on their points.  Making these instructions act like memcpy provides maximal flexibility and there does not seem to be any significant downside to that.

jfbastien:
The instruction would just be a "copy to memory from read-only segment"? Won't really have anything to do with initialization anymore. That's kinda neat. I assume we'll spec the writes as non-atomic, ordered from low to high bits?

binji:
Addressed some of the comments, is this clear enough or should I go into more detail?

> I assume we'll spec the writes as non-atomic, ordered from low to high bits?

@jfbastien Yes, we already have similar wording [here](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#instantiation), I imagine we'll just have a spec "function" that shares that behavior.

proposals/threads/ConditionalSegmentInitialization.md
```diff
@@ -175,3 +175,75 @@ elem    ::= 0x01 e_o:expr e_a:expr y*:vec(funcindex) => {table 0, offset e_o, ap
...
+offset into the target memory, and a length to copy.
+
+When `init_memory` is executed, its behavior matches the steps described in
+step 11 of
+[instantiation](https://webassembly.github.io/spec/exec/modules.html#instantiation),
+but it behaves as though the segment were specified with the source offset,
+target offset, and length as given by the `init_memory` operands.
+
+A trap occurs if any of the accessed bytes lies outside the source data segment
+or the target memory.
```
lukewagner: I think the "trap if not during the call-start-function phase of instantiation" bit seems to have gone away.

jfbastien:
Why is "trap if not during the call-start-function phase of instantiation" desirable?

lukewagner:
See [comment in #62](https://github.com/WebAssembly/threads/issues/62#issuecomment-342545564) and following.

jfbastien:
> See [comment in #62](https://github.com/WebAssembly/threads/issues/62#issuecomment-342545564) and following.

That seems like an odd restriction from a user perspective, and is odder to implement than the alternative. Even the optimization where we throw away now-dead segments isn't that hard if `start` is the only caller and isn't exported (call graph not needed, just "is called" bit on each function).

lukewagner:
Yes, but that means there will be an implicit memory usage cliff that could be triggered by something as arbitrary as your C++ compiler's inlining heuristics.  The implementation is quite trivial (it's certainly more work to do heuristics that best-effort discard segments) and the restriction isn't odd when the performance situation is explained (as is the case for much of wasm).

rossberg:
Agreed it is kind of odd, but it is conservative -- we could lift the restriction later. Alternatively, we could extend the segment flag to explicitly indicate whether a given segment should be usable after initialisation.

jfbastien:
I think we should explicitly discuss this at an upcoming meeting.

PR #81, opened 2017-11-16
Update docs to allow import/export mut globals

binji:


rossberg:
Looks good, but please remove the respective notes in syntax/modules.rst as well.

binji:
Thanks, didn't notice those at first!

PR #83, opened 2017-11-27
Fix code formatting

sbc100:


binji:
It's weird, vim always wants me to escape underscores in code blocks, but I guess GH doesn't really care.

sbc100:
I think your vim is wrong.   My vim doesn't ask me to do that.   If you do it produces a literal backslash in the rendered output (at least on github)

binji:
Oh, you're right. It's not in code blocks, it complains about it elsewhere. Which makes sense since `_foo_` is used for italics. _foo_

PR #85, opened 2017-12-18
Update proposal to include {mem,table}.drop

binji:


binji:
See https://github.com/WebAssembly/threads/issues/62#issuecomment-352185562.

proposals/threads/ConditionalSegmentInitialization.md
```diff
@@ -74,17 +74,29 @@ requires two modules where one should be enough.
...
+
+* `mem.init`: copy a region from a data segment
+* `table.init`: copy an region from an element segment
 
 An inactive segment has no initializer expression, since it will be specified
-as an operand to `init_memory` or `init_table`.
+as an operand to `mem.init` or `table.init`.
+
+Segments (either active or inactive) can also be discarded by using the
+following new instructions:
```
lars-t-hansen: Gosh, I'm not sure if it's a good idea to be able to drop active memories.  Suppose you can, and do.  Then you instantiate the module again.  What happens then?  A trap?  (That would be the interpretation of the prose further down.)

binji: Yeah, I wasn't sure about this one. I was thinking it would be a nice optimization to be able to discard to active segments (say, if you know that there will never be another instance), but perhaps we should be more conservative.

rossberg: I'd err on the side of conservative.

binji: OK, changed.

proposals/threads/ConditionalSegmentInitialization.md
```diff
@@ -96,36 +109,41 @@ data    ::= 0x01 b*:vec(byte)            => {data 0, offset empty, init b*, acti
...
+target offset, and length as given by the `mem.init` operands.
 
 A trap occurs if any of the accessed bytes lies outside the source data segment
-or the target memory.
+or the target memory. A trap also occurs if the segment is used after it has
+been dropped via `mem.drop`.
 
-Note that it is allowed to use `init_memory` on the same data segment more than
+Note that it is allowed to use `mem.init` on the same data segment more than
 once, or with an active data segment.
```
lars-t-hansen: I think it's a mistake to allow mem.init to be used with an active data segment.  The current contract is that the active data segment may be dropped by the implementation once the Module object becomes GC-able.  This makes mem.init sensitive to GC timing, because it'll have to trap if the data segment has been removed.

IMO all the programmatic mem opcodes should only be usable with the inactive data segments.

(I also have decided I dislike the terminology "active" and "inactive", but I've been at a loss to propose something better.  Still am, really.  But "instantiation-time" vs "run-time" would capture the distinction better.)

binji: Yes, I think you're right. I'll try to find some time to update before the end of the year... (at home with the family :smile:).

binji:
I've renamed `inactive` to `passive` and also made it a validation error to use `*.init` or `*.drop` with active segments.

binji:
Landing for now, we can iterate in further PRs.

PR #90, opened 2018-03-04
Define postMessage of shared Memory

littledan:
[spec] Normative: Store is per agent cluster

With the WebAssembly threads proposal, a single Memory can be shared between
multiple agents within an agent cluster. This patch associates a WebAssembly
store with an agent cluster, rather than an agent, and updates the WebAssembly
JS API to define the store for an agent as the store for its surrounding
agent cluster.

[spec] Normative: Define serialization for SAB Memory

The semantics here are
- forStorage serialization is disallowed
- Transfers are only permitted within an agent cluster
- Transfer of Memory backed by SharedArrayBuffers makes the buffer available in
  the source and the destination, with no resulting detached buffer.
- Transfer of Memory backed by ArrayBuffers is disallowed.

This specification attempts to be well-integrated with HTML by defining
n terms of the serialization behavior of Memory as a platform object, and
using the sub-serialization of the underlying SharedArrayBuffer to check
for transferring within an agent cluster, and not to IndexedDB.

Based on the earlier PR at WebAssembly/design#1074

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -212,9 +212,7 @@ fetch('demo.wasm').then(response =>
 
 Note: WebAssembly semantics are defined in terms of an abstract [=store=], representing the state of the WebAssembly abstract machine. WebAssembly operations take a store and return an updated store.
 
-Each [=agent=] has an <dfn>associated store</dfn>. When a new agent is created, its associated store is set to the result of [=init_store=]().
-
-Note: In this specification, no WebAssembly-related objects, memory or addresses can be shared among agents in an [=agent cluster=]. In a future version of WebAssembly, this may change.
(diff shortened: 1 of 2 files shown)


littledan:
This patch is missing tests; we should follow up with those. But it looks like there are a lot of missing tests compared to the current specification, so maybe this shouldn't block landing. Feel free to merge this patch if you feel it's appropriate; I don't have any other planned changes.

lars-t-hansen:
LGTM too.  There's an implementation of (what purports to be) this algorithm in Firefox Nightly.  I do have some test cases - using the FF test harness - in https://dxr.mozilla.org/mozilla-central/source/js/src/jit-test/tests/wasm/memory-*.js.

littledan:
@lars-t-hansen Are you facing any sort of barrier to developing these as web-platform-tests rather than using the Firefox test harness?

lars-t-hansen:
It's a lame excuse, but I can write jit-tests in my sleep while wpt requires a larger time investment...  Also the jit-tests run in the spidermonkey shell, which is important for development speed, I don't want to rebuild Firefox in order to test.

binji:
Cool, I'm OK with landing tests in another PR.

Issue #92, opened 2018-04-19
Alignment of wait/wake


conrad-watt:
The current overview has this to say about wait/wake alignment:

"Both wake and wait operators trap if the effective address of either operator is misaligned or out-of-bounds. The wait operators requires an alignment of their memory access size. The wait operator requires an alignment of 32 bits."

There seem to be some minor typos here, and it's worth clarifying: are we allowing 64 bit wait/wake to be "only" 32 bit aligned?

lars-t-hansen:
Yes, that looks buggy + outdated.  We require natural alignment for 64-bit.

lars-t-hansen:
I take back the bit about the alignment on wake.  There was a discussion.  We decided to require that wake only has 32-bit alignment.  The reason for that is that wake does not need to inspect the memory at the address it's given, and 32-bit alignment is compatible with JS, thus improving / ensuring interop.

conrad-watt:
Does i64.wake have any semantic distinction from i32.wake in this case?

If I do an i64.wake across the edge of memory (i.e. beginning at the last mapped 32 bits), should this trap?

lars-t-hansen:
> Does i64.wake have any semantic distinction from i32.wake in this case?

There are no 'i32.wake' and 'i64.wake' operations, there is just 'wake', which takes an address, which must be 32-bit aligned for compatibility with JS and the i32.wait/i64.wait operations.  At least that was the case when I implemented this, and it matches my memory about discussions, and I don't think this should have changed since.

The address argument to wake is checked for being oob but at least in Firefox we assume the width is just 1 byte.

conrad-watt:
Ah - my misunderstanding. I see there's only one wake op, That all makes sense now. Thanks!

PR #93, opened 2018-04-19
Fix minor typos in wait/wake proposal

conrad-watt:
Following on from discussion in #92. Wake should be 32-bit aligned, not wait.


binji:
Thanks!

PR #102, opened 2018-08-21
[spec] First steps towards an abstract machine with threads & events

rossberg:
This PR:
* Define and describe actions and events.
* Change memory instances to externalise state.
* Change addresses into abstract values.
* Change configurations to include traces and multiple threads.
* Add an item on threads to the overview.

TODO:
* Change store representation from lists to maps.
* Externalise state of tables and globals for uniformity.
* Change definition of host functions.
* Change definitions of evaluation contexts.
* Change reduction rules to use actions.
* Define axiomatic ordering constraints.

### DIFF
--- a/document/core/exec/runtime.rst
+++ b/document/core/exec/runtime.rst
@@ -59,6 +59,8 @@ It is either a sequence of :ref:`values <syntax-val>` or a :ref:`trap <syntax-tr
 Store
 ~~~~~
 
+TODO: since allocation order is non-deterministic in the presence of threads, we can no longer use ordered lists but need mappings from abstract addresses to instances.
+
 The *store* represents all global state that can be manipulated by WebAssembly programs.
 It consists of the runtime representation of all *instances* of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tableinst>`, :ref:`memories <syntax-meminst>`, and :ref:`globals <syntax-globalinst>` that have been :ref:`allocated <alloc>` during the life time of the abstract machine. [#gc]_
 
@@ -106,12 +108,13 @@ Addresses
 ~~~~~~~~~
 
 :ref:`Function instances <syntax-funcinst>`, :ref:`table instances <syntax-tableinst>`, :ref:`memory instances <syntax-meminst>`, and :ref:`global instances <syntax-globalinst>` in the :ref:`store <syntax-store>` are referenced with abstract *addresses*.
-These are simply indices into the respective store component.
+Each address uniquely determines a respective component in the store.
+Other than that the form that addresses take is unspecified and cannot be observed.
 
 .. math::
    \begin{array}{llll}
    \production{(address)} & \addr &::=&
-     0 ~|~ 1 ~|~ 2 ~|~ \dots \\
+     \dots \\
    \production{(function address)} & \funcaddr &::=&
      \addr \\
    \production{(table address)} & \tableaddr &::=&
@@ -132,9 +135,6 @@ even where this identity is not observable from within WebAssembly code itself
    which are *static*, module-local references to their original definitions.
    A *memory address* |memaddr| denotes the abstract address *of* a memory *instance* in the store,
    not an offset *inside* a memory instance.
-
-   There is no specific limit on the number of allocations of store objects,
-   hence logical addresses can be arbitrarily large natural numbers.
 
 
 .. index:: ! instance, function type, function instance, table instance, memory instance, global instance, export instance, table address, memory address, global address, index, name
@@ -189,6 +189,8 @@ The module instance is used to resolve references to other definitions during ex
    \production{(host function)} & \hostfunc &::=& \dots \\
    \end{array}
 
+TODO: need to represent host functions differently to encompass threading
+
 A *host function* is a function expressed outside WebAssembly but passed to a :ref:`module <syntax-module>` as an :ref:`import <syntax-import>`.
 The definition and behavior of host functions are outside the scope of this specification.
 For the purpose of this specification, it is assumed that when :ref:`invoked <exec-invoke-host>`,
@@ -208,6 +210,8 @@ but within certain :ref:`constraints <exec-invoke-host>` that ensure the integri
 
 Table Instances
 ~~~~~~~~~~~~~~~
+
+TODO: table access should uniformly happen through events now
 
 A *table instance* is the runtime representation of a :ref:`table <syntax-table>`.
 It holds a vector of *function elements* and an optional maximum size, if one was specified in the :ref:`table type <syntax-tabletype>` at the table's definition site.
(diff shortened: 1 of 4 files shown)


rossberg:
@conrad-watt, @binji, PTAL.

document/core/exec/runtime.rst
```diff
@@ -523,37 +642,37 @@ This definition allows to index active labels surrounding a :ref:`branch <syntax
...
+.. index:: ! configuration, ! thread, ! trace, store, event, frame, instruction, module instruction
 .. _syntax-thread:
+.. _syntax-trace:
 .. _syntax-config:
 
 Configurations
 ..............
 
-A *configuration* consists of the current :ref:`store <syntax-store>` and an executing *thread*.
+A *configuration* consists of the current :ref:`store <syntax-store>`, a *trace* of the store :ref:`events <syntax-evt>` that have occurred so far, and a set of executing *threads*.
```
conrad-watt: Depending on how we choose to define valid executions, we may end up deciding not to keep track of the events so far at the operational level. It's probably fine for now though.

document/core/exec/runtime.rst
```diff
@@ -330,6 +356,99 @@ It filters out entries of a specific kind in an order-preserving fashion:
...
+   Future versions of WebAssembly may introduce additional orderings.
+
+A |AWAKE| action is performed by the |ATOMICWAKE| instruction.
+It records the :ref:`address <syntax-memaddr>` of the :ref:`memory instance <syntax-meminst>`, the location, and the number of suspended threads woken up.
+
+Finally, a *host action* is an action performed outside of WebAssembly code.
+Its form and meaning is outside the scope of this specification.
+
+.. note::
+   An :ref:`embedder <embedder>` may define a custom set of host actions and respective ordering constraints to model other forms of interactions that are not expressible withing WebAssembly, but whose ordering relative to WebAssembly events is relevant for the combined semantics.
```
conrad-watt: Nit: "withing WebAssembly"

document/core/exec/runtime.rst
```diff
@@ -239,20 +243,18 @@ Memory Instances
...
+Like in a :ref:`memory type <syntax-memtype>`, the maximum size in a memory instance is given in units of the WebAssembly *page size*, which is defined to be the constant :math:`65536` -- abbreviated :math:`64\,\F{Ki}`.
 
-The bytes can be mutated through :ref:`memory instructions <syntax-instr-memory>`, the execution of a :ref:`data segment <syntax-data>`, or by external means provided by the :ref:`embedder <embedder>`.
-
-It is an invariant of the semantics that the length of the byte vector, divided by page size, never exceeds the maximum size, if present.
+WebAssembly :ref:`threads <thread>` expose a *weak memory model* in which reads, writes, and other accesses to a memory instance can be reordered in a non-deterministic fashion that mirrors the behavior of hardware.
+Instead of representing the contents of a memory directly in an instance -- e.g., as a specific vector of :ref:`bytes <syntax-byte>` -- the abstract machine hence separately records :ref:`traces <syntax-trace>` of corresponding memory :ref:`events <syntax-evt>` that describe the history of all accesses that have occured.
```
conrad-watt: If we go with the coinductive definition, a trace won't keep track of the events that have occurred, but all sequences of events that _can_ occur, starting from a configuration. Still to be discussed properly though (see my other comment below).

document/core/exec/runtime.rst
```diff
@@ -330,6 +356,99 @@ It filters out entries of a specific kind in an order-preserving fashion:
...
+The interaction of a WebAssembly computation with the :ref:`store <syntax-store>` is described through *actions*, such as reads and writes.
+The execution of every individual :ref:`instruction <syntax-instr>` can perform a set of such actions.
+
+.. math::
+   \begin{array}{llcl}
+   \production{(action)} & \act &::=&
+     \AFORK~\instr^\ast \\&&|&
+     \ANEW~\addr~\externinst \\&&|&
+     \AUSE~\addr~\externinst \\&&|&
+     \ARD_\ord~\addr~\loc~b^\ast \\&&|&
```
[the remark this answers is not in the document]
conrad-watt: also same with write/rmw below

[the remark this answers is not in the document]
rossberg: Reading \K{len} will be a u32 (later u64), reading a global an arbitrary value. So it seemed to make sense to simply include values. I don't think anything more exotic will ever be needed, but one can always extend the definition.

document/core/exec/runtime.rst
```diff
@@ -330,6 +356,99 @@ It filters out entries of a specific kind in an order-preserving fashion:
...
+     \AWR_\ord~\addr~\loc~b^\ast \\&&|&
+     \ARMW~\addr~\loc~b^\ast~b^ast \\&&|&
+     \AWAKE~\addr~\loc~\u32 \\&&|&
```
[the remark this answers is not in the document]
rossberg: I think you're right. I believe we could simplify to a Boolean, but there probably isn't much point. Changed.

document/core/exec/runtime.rst
```diff
@@ -330,6 +356,99 @@ It filters out entries of a specific kind in an order-preserving fashion:
...
+     \LVAL \\
+   \production{(store value)} & \storeval &::=&
+     \val ~|~
+     b^\ast \\
+   \end{array}
+
+Performing as |AFORK| action creates a new :ref:`thread <thread>` of computation, as given by a sequence of :ref:`instructions <syntax-instr>`.
+
+The |ANEW| action creates a new :ref:`external instance <syntax-externinst>` at a fresh :ref:`address <syntax-addr>`.
+The corresponding |AUSE| action accesses a previously created external instance through its address.
```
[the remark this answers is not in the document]
rossberg: Yeah, not the perfect name, but it's consistent with the existing notions of external value and external type. Not sure we want to change those.

document/core/exec/runtime.rst
rossberg: Fixed.

document/core/exec/runtime.rst
rossberg: Yes, would have to think much harder how to word this otherwise.

document/core/exec/runtime.rst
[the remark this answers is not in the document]
rossberg: Oops, now I realise that you were pointing out the b^\ast that should be \storeval. Fixed.

document/core/exec/runtime.rst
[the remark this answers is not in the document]
conrad-watt: I suppose it's fine given that. Maybe we'll have an inspiration later.

conrad-watt:
> Since we generally need to refactor the way we access the store it makes sense to use the labeled transition system consistently.

@rossberg just to check I'm not misunderstanding anything, what were you referring to when you wrote this?

document/core/exec/runtime.rst
binji: sp: occurred

document/core/exec/runtime.rst
```diff
@@ -330,6 +356,99 @@ It filters out entries of a specific kind in an order-preserving fashion:
...
+   \production{(location)} & \loc &::=&
+     \u32 ~|~
+     \LSIZE ~|~
+     \LVAL \\
+   \production{(store value)} & \storeval &::=&
+     \val ~|~
+     b^\ast \\
+   \end{array}
+
+Performing as |AFORK| action creates a new :ref:`thread <thread>` of computation, as given by a sequence of :ref:`instructions <syntax-instr>`.
```
binji: s/Performing as/Performing a/

document/core/exec/runtime.rst
```diff
@@ -330,6 +356,99 @@ It filters out entries of a specific kind in an order-preserving fashion:
...
+
+.. math::
+   \begin{array}{llcl}
+   \production{(action)} & \act &::=&
+     \AFORK~\instr^\ast \\&&|&
+     \ANEW~\addr~\externinst \\&&|&
+     \AUSE~\addr~\externinst \\&&|&
+     \ARD_\ord~\addr~\loc~\storeval \\&&|&
+     \AWR_\ord~\addr~\loc~\storeval \\&&|&
+     \ARMW~\addr~\loc~\storeval~\storeval \\&&|&
```
binji: some way to annotate which storeval is read and which is written?

PR #114, opened 2019-01-15
[test] Add js-api tests

binji:
This tests the `WebAssembly.Memory` constructor and the `grow` method.

### DIFF
--- a/test/js-api/memory/constructor.any.js
+++ b/test/js-api/memory/constructor.any.js
@@ -8,16 +8,18 @@ function assert_Memory(memory, expected) {
 
   // https://github.com/WebAssembly/spec/issues/840
   assert_equals(memory.buffer, memory.buffer, "buffer should be idempotent");
-  const bufferType = expected.shared ? SharedArrayBuffer : ArrayBuffer;
+  const isShared = !!expected.shared;
+  const bufferType = isShared ? SharedArrayBuffer : ArrayBuffer;
   assert_equals(Object.getPrototypeOf(memory.buffer), bufferType.prototype,
                 'prototype of buffer');
-  assert_true(Object.isExtensible(memory.buffer), "buffer extensibility");
   assert_equals(memory.buffer.byteLength, 0x10000 * expected.size, "size of buffer");
   if (expected.size > 0) {
     const array = new Uint8Array(memory.buffer);
     assert_equals(array[0], 0, "first element of buffer");
     assert_equals(array[array.byteLength - 1], 0, "last element of buffer");
   }
+  assert_equals(isShared, Object.isFrozen(memory.buffer), "buffer frozen");
+  assert_not_equals(isShared, Object.isExtensible(memory.buffer), "buffer extensibility");
 }
 
 test(() => {
--- a/test/js-api/memory/constructor.any.js
+++ b/test/js-api/memory/constructor.any.js
@@ -8,8 +8,9 @@ function assert_Memory(memory, expected) {
 
   // https://github.com/WebAssembly/spec/issues/840
   assert_equals(memory.buffer, memory.buffer, "buffer should be idempotent");
-  assert_equals(Object.getPrototypeOf(memory.buffer), ArrayBuffer.prototype,
-                "prototype of buffer");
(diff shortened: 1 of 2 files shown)


binji:
I haven't tested this yet against an implementation, so it's probably broken :-}

binji:
cc @Ms2ger (can't add you as a reviewer)

test/js-api/memory/constructor.any.js
```diff
@@ -8,15 +8,18 @@ function assert_Memory(memory, expected) {
 
   // https://github.com/WebAssembly/spec/issues/840
   assert_equals(memory.buffer, memory.buffer, "buffer should be idempotent");
-  assert_equals(Object.getPrototypeOf(memory.buffer), ArrayBuffer.prototype,
-                "prototype of buffer");
-  assert_true(Object.isExtensible(memory.buffer), "buffer extensibility");
+  const isShared = !!expected.shared;
+  const bufferType = isShared ? SharedArrayBuffer : ArrayBuffer;
```
Ms2ger: I think this is going to cause a `ReferenceError` if `SharedArrayBuffer` is not present, even in the not-shared case. Can you use `self.SharedArrayBuffer` instead?

test/js-api/memory/constructor.any.js
```diff
@@ -8,15 +8,18 @@ function assert_Memory(memory, expected) {
...
+  assert_equals(Object.getPrototypeOf(memory.buffer), bufferType.prototype,
+                'prototype of buffer');
   assert_equals(memory.buffer.byteLength, 0x10000 * expected.size, "size of buffer");
   if (expected.size > 0) {
     const array = new Uint8Array(memory.buffer);
     assert_equals(array[0], 0, "first element of buffer");
     assert_equals(array[array.byteLength - 1], 0, "last element of buffer");
   }
+  assert_equals(isShared, Object.isFrozen(memory.buffer), "buffer frozen");
+  assert_not_equals(isShared, Object.isExtensible(memory.buffer), "buffer extensibility");
```
Ms2ger: Nit: swap the order of the arguments (actual first, expected second).

test/js-api/memory/constructor.any.js
```diff
@@ -118,13 +125,25 @@ test(() => {
         },
       };
     },
+
+    get shared() {
+      order.push("shared");
+      return {
+        valueOf() {
+          order.push("shared valueOf");
+          return 1;
```
Ms2ger: Maybe `true`

test/js-api/memory/grow.any.js
```diff
@@ -2,7 +2,8 @@
 
 function assert_ArrayBuffer(actual, expected, message) {
   // https://github.com/WebAssembly/spec/issues/840
-  assert_equals(Object.getPrototypeOf(actual), ArrayBuffer.prototype,
+  const bufferType = expected.shared ? SharedArrayBuffer : ArrayBuffer;
```
Ms2ger: Same question

test/js-api/memory/grow.any.js
```diff
@@ -183,3 +184,29 @@ test(() => {
...
+  const memory = new WebAssembly.Memory(argument);
+  const oldMemory = memory.buffer;
+  assert_ArrayBuffer(oldMemory, { "size": 1, "shared": true }, "Buffer before growing");
+
+  const result = memory.grow(1);
+  assert_equals(result, 1);
+
+  const newMemory = memory.buffer;
+  assert_not_equals(oldMemory, newMemory);
+  assert_ArrayBuffer(oldMemory, { "size": 1, "shared": true }, "Old buffer after growing");
```
Ms2ger: Why is this not detached?

binji: SharedArrayBuffers can't be detached.

binji:
> Looks like you still need to write the spec text for all of this?

Yes, I should probably have done that first. :-) But an old version of some of the changes is in the [overview](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#javascript-api-changes).

PR #115, opened 2019-01-16
[interpreter] Add *.atomic.wait and atomic.notify

binji:
These are not implemented for execution.

### DIFF
--- a/interpreter/syntax/ast.ml
+++ b/interpreter/syntax/ast.ml
@@ -62,10 +62,14 @@ type 'a memop =
   {ty : value_type; align : int; offset : Memory.offset; sz : 'a option}
 type loadop = (Memory.pack_size * Memory.extension) memop
 type storeop = Memory.pack_size memop
-type atomicloadop = storeop (* Memory.extension isn't used. *)
-type atomicstoreop = storeop
-type atomicrmwop = rmwop * storeop
-type atomicrmwcmpxchgop = storeop
+
+type atomicmemop = storeop (* No atomic operations use Memory.extension *)
+type atomicwaitop = atomicmemop
+type atomicnotifyop = atomicmemop
+type atomicloadop = atomicmemop
+type atomicstoreop = atomicmemop
+type atomicrmwop = rmwop * atomicmemop
+type atomicrmwcmpxchgop = atomicmemop
 
 
 (* Expressions *)
@@ -104,6 +108,8 @@ and instr' =
   | Unary of unop                     (* unary numeric operator *)
   | Binary of binop                   (* binary numeric operator *)
   | Convert of cvtop                  (* conversion *)
+  | AtomicWait of atomicwaitop        (* atomically wait for notification at address *)
+  | AtomicNotify of atomicnotifyop    (* atomically notify all waiters at address *)
   | AtomicLoad of atomicloadop        (* atomically read memory at address *)
   | AtomicStore of atomicstoreop      (* atomically write memory at address *)
   | AtomicRmw of atomicrmwop          (* atomically read, modify, write memory at address *)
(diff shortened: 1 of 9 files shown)


interpreter/syntax/ast.ml
```diff
@@ -62,10 +62,14 @@ type 'a memop =
   {ty : value_type; align : int; offset : Memory.offset; sz : 'a option}
 type loadop = (Memory.pack_size * Memory.extension) memop
 type storeop = Memory.pack_size memop
-type atomicloadop = storeop (* Memory.extension isn't used. *)
-type atomicstoreop = storeop
-type atomicrmwop = rmwop * storeop
-type atomicrmwcmpxchgop = storeop
+
+type atomicmemop = storeop (* No atomic operations use Memory.extension *)
```
rossberg: I would perhaps remove all the aliases to this type (and rename this to just `atomicop`?); `atomicrmwop` could go as well -- it seems more logical to have the rmwop as a separate argument directly on the `AtomicRMW` constructor, since it isn't really related to the memory access per se.

interpreter/text/arrange.ml
```diff
@@ -205,11 +205,14 @@ let rmw = function
   | I32 I32Op.RmwXchg | I64 I64Op.RmwXchg -> "xchg"
   | _ -> assert false
 
-let memop name {ty; align; offset; _} =
-  value_type ty ^ "." ^ name ^
+let memarg {ty; align; offset; _} =
```
rossberg: Nit: since this one outputs a proper memop, I'd rather keep its name, and call the new one something different (`memop_without_type`?). Also, I think you can keep the name argument, since even the new function prepends it.

test/core/atomic.wast
```diff
@@ -464,3 +464,16 @@
 (assert_invalid (module (memory 1 1) (func (drop (i32.atomic.rmw16_u.cmpxchg (i32.const 0) (i32.const 0) (i32.const 0))))) "atomic accesses require shared memory")
 (assert_invalid (module (memory 1 1) (func (drop (i64.atomic.rmw16_u.cmpxchg (i32.const 0) (i64.const 0) (i64.const 0))))) "atomic accesses require shared memory")
 (assert_invalid (module (memory 1 1) (func (drop (i64.atomic.rmw32_u.cmpxchg (i32.const 0) (i64.const 0) (i64.const 0))))) "atomic accesses require shared memory")
+
+;; *.atomic.wait and atomic.notify (unimplemented in exec)
```
rossberg: Make this a TODO?

interpreter/syntax/ast.ml
binji: Done

interpreter/text/arrange.ml
binji: Done

test/core/atomic.wast
binji: Done

interpreter/text/arrange.ml
```diff
@@ -205,11 +205,15 @@ let rmw = function
   | I32 I32Op.RmwXchg | I64 I64Op.RmwXchg -> "xchg"
   | _ -> assert false
 
-let memop name {ty; align; offset; _} =
-  value_type ty ^ "." ^ name ^
+let memop_without_type name {ty; align; offset; _} =
```
rossberg: Nit: can remove `ty` in pattern.

binji: `ty` is used below when writing alignment. Maybe the name makes this unclear -- it could be `memop_without_type_prefix` perhaps?

rossberg: Ah, sorry, missed that. I think it's fine as is then.

PR #116, opened 2019-01-23
Add JS API spec text

binji:
In the overview, the `Memory.buffer` getter only creates a new
`SharedArrayBuffer` when the lengths don't match. In the JS spec, there
is a Memory object cache which is unconditionally updated with the most
recent memory object and its length after every grow operation. So I
think it makes sense to specify that the `buffer` getter always accesses
this cache, since it is simpler (and unobservable, I think).

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -28,6 +28,7 @@ Prepare For TR: true
 urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
     type: interface; for: ECMAScript
         text: ArrayBuffer; url: sec-arraybuffer-objects
+        text: SharedArrayBuffer; url: sec-sharedarraybuffer-objects
     type: exception; for: ECMAScript
         text: Error; url: sec-error-objects
         text: NativeError; url: sec-nativeerror-constructors
@@ -36,7 +37,9 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
     type: dfn
         text: agent cluster; url: sec-agent-clusters
         text: agent; url: agent
-        text: data block; url: sec-data-blocks
+        url: sec-data-blocks
+            text: data block
+            text: shared data block
         text: Bound Function; url: sec-bound-function-exotic-objects
         text: NumericLiteral; url: sec-literals-numeric-literals
         text: surrounding agent; url: surrounding-agent
@@ -537,12 +540,13 @@ interface Instance {
 dictionary MemoryDescriptor {
   required [EnforceRange] unsigned long initial;
   [EnforceRange] unsigned long maximum;
+  boolean shared = false;
 };
 
 [LegacyNamespace=WebAssembly, Constructor(MemoryDescriptor descriptor), Exposed=(Window,Worker,Worklet)]
 interface Memory {
   unsigned long grow([EnforceRange] unsigned long delta);
-  readonly attribute ArrayBuffer buffer;
+  readonly attribute (ArrayBuffer or SharedArrayBuffer) buffer;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


binji:
cc @Ms2ger

document/js-api/index.bs
```diff
@@ -623,7 +634,21 @@ Immediately after a WebAssembly [=memory.grow=] instruction executes, perform th
 </div>
 
 <div algorithm>
-    The getter of the <dfn attribute for="Memory">buffer</dfn> attribute of {{Memory}} returns the receiver's \[[BufferObject]] internal slot.
+    The getter of the <dfn attribute for="Memory">buffer</dfn> attribute of {{Memory}} perfoms the following steps:
+
+    1. Let |memory| be the Memory instance.
+    1. Let |memaddr| be |memory|.\[[Memory]].
+    1. Let |block| be a [=Data Block=] which is [=identified with=] the underlying memory of |memaddr|.
+    1. If |block| is a [=Data Block=],
```
Ms2ger: This is kinda strange; a "Shared Data Block" has to be and not be a "Data Block" in these two steps.

document/js-api/index.bs
```diff
@@ -623,7 +634,21 @@ Immediately after a WebAssembly [=memory.grow=] instruction executes, perform th
...
+        1. Return |memory|.\[[BufferObject]].
+    1. Otherwise,
+        1. Assert: |block| is a [=Shared Data Block=].
+        1. Let |map| be the [=surrounding agent=]'s associated [=Memory object cache=].
+        1. Assert: |map|[|memaddr|] [=map/exists=].
+        1. Let |newMemory| be |map|[|memaddr|].
```
Ms2ger: I don't understand what's going on here. It seems like `memory` and `newMemory` should always be the same object?

document/js-api/index.bs
xtuc: I would prefer relying on the shared attibute.

binji: It's weird, but I couldn't really see another way to distinguish a data block [in the ECMAScript spec](https://tc39.github.io/ecma262/#sec-data-blocks):

> A data block that resides in memory that can be referenced from multiple agents concurrently is designated a Shared Data Block. ... Finally, Shared Data Blocks can be distinguished from Data Blocks.

> I would prefer relying on the shared attibute.

That feels a bit more indirect to me, but seems like it would work too.

document/js-api/index.bs
binji: It's possible to have two memory objects that are associated with the same data block, for example if a `Memory` object is postMessage'd to a worker and back. The one that grew was written to the cache, but the other is not.

document/js-api/index.bs
littledan: @xtuc "shared" is just part of the dictionary passed into the constructor, and it's not present at this point.

Another way to phrase things would be to reverse the two conditionals, so it's "If block is a Shared Data Block/otherwise". That might be more clear.

document/js-api/index.bs
littledan: Do you have any Web API spec text drafted for how serialization works for Memory?

binji: Not yet, I'll work on that next.

document/js-api/index.bs
binji: > "shared" is just part of the dictionary passed into the constructor, and it's not present at this point.

Oh, I assumed @xtuc meant the shared attribute of the wasm-side memory instance.

binji: > Another way to phrase things would be to reverse the two conditionals, so it's "If block is a Shared Data Block/otherwise". That might be more clear.

Done.

document/js-api/index.bs
binji: Oh, you already did this here: https://github.com/WebAssembly/threads/pull/90. Do we need to update this at all?

littledan: Haha I forgot! Well, this change makes @Ms2ger 's question clear: It will actually be the *same* memory object! In particular, when the shared Memory object is postMessaged back, the original Memory object will be found in the cache.

binji: Oh, right. So this actually ends up being much simpler then.

PR #117, opened 2019-01-23
Spec text/formal: atomic on non-shared memory is a validation error

conrad-watt:
As indicated in the [overview](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#atomic-memory-accesses), "attempting to use atomic access operators on non-shared linear memory is a validation error".

Question: is "shareability" the right informal name for the _shared_ portion of the memory type? I debated "sharedness".


rossberg:
I agree that "sharedness" may sound more fitting, but I'm not a native speaker. :)

PR #119, opened 2019-01-30
[interp] Add command to spawn threads

binji:
This command is meant to execute each action on a separate thread, then
validate that each returns the expected value.

It's currently unimplemented in both `script/run.ml` and `script/js.ml`.

### DIFF
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -312,7 +312,7 @@ cmd:
 module with given failure string
   <action>                                   ;; perform action and print results
   <assertion>                                ;; assert result of an action
-  ( threads <assertion>* )                   ;; run all assertions in parallel
+  ( spawn <name>?> action )                  ;; spawn a new thread and run an action
   <meta>                                     ;; meta command
 
 module:
@@ -323,6 +323,7 @@ module:
 action:
   ( invoke <name>? <string> <expr>* )        ;; invoke function export
   ( get <name>? <string> )                   ;; get global export
+  ( join <name> )                            ;; join named thread and return its value
 
 assertion:
   ( assert_return <action> <expr>* )         ;; assert action has expected results
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -312,6 +312,7 @@ cmd:
 module with given failure string
   <action>                                   ;; perform action and print results
   <assertion>                                ;; assert result of an action
+  ( threads <assertion>* )                   ;; run all assertions in parallel
   <meta>                                     ;; meta command
 
 module:
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -312,7 +312,7 @@ cmd:
 module with given failure string
   <action>                                   ;; perform action and print results
   <assertion>                                ;; assert result of an action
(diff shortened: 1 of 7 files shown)


rossberg:
Can you clarify what the intended use of this feature is? Is this to write tests involving concurrency or is it to parallelise tests?

Also, is there a plan how to implement this? It isn't entirely obvious to me how to do this in the interpreter without using actual threads, but for the spec interpreter we rather want a literal implementation of the interleaved small-step reduction rules.

I see that for concurrency tests we somehow need to be able to express spawning threads. As long as Wasm itself does not support that I would suggest adding a spawn construct on the action level of the script language, which corresponds more closely to "the host". Assertions would remain as they are.

binji:
The purpose was to have a high-level way to express tests that use threads in a race-free way. Something like:

```wasm
(module
  (func (export "wait") (result i32) ... <wait for notification> (i32.const 1))
  (func (export "notify") (result i32) ... <notify> (i32.const 2)))

(threads
  (assert_return (invoke "wait") (i32.const 1))
  (assert_return (invoke "notify") (i32.const 2)))
```

(ignore that this is racy, you get the idea :-))

I thought about adding `spawn`, but I'd guess we'd need a `join` as well to get the result, and a way to specify the thread to join:

```wasm
<same module>

(spawn $wait "wait")
(spawn $notify "notify")
(assert_return (join $wait) (i32.const 1))
(assert_return (join $notify) (i32.const 2))
```

I think this is OK. It's a bit more verbose, but also more powerful.

> Also, is there a plan how to implement this? It isn't entirely obvious to me how to do this in the interpreter without using actual threads, but for the spec interpreter we rather want a literal implementation of the interleaved small-step reduction rules.

No, I'm not certain how we'd want to do this, though I agree the spec interpreter will want to follow the spec as closely as possible. I'm not quite sure how we handle the non-determinism however.

My main interest in adding this feature is for writing tests that can be translated to JS so it can be tested by the browser VMs.

binji:
Updated to use `spawn` and `join` instead.

interpreter/README.md
```diff
@@ -312,6 +312,7 @@ cmd:
 module with given failure string
   <action>                                   ;; perform action and print results
   <assertion>                                ;; assert result of an action
+  ( spawn <name>?> action )                  ;; spawn a new thread and run an action
```
rossberg: Stray `>`; also `<action>`

binji:
Landing this part for now, still working on JS part for a separate PR.

binji:
> Should a thread perhaps be allowed to have multiple actions?
Since `spawn` is a binder for a thread name now, maybe `thread` is still the better keyword?

Oops, didn't address these. I think it could have multiple actions, but I'm having enough trouble w/ one action for now. :-)
`thread` WFM, will change in another PR.

rossberg:
Hm, I don't have any good suggestion to offer regarding the worker issue. The script language predates threads, and didn't distinguish between defining and instantiating a module. We should probably have a way to define memories/tables/globals outside modules, to emulate the host API, as well as registering them as imports.

PR #126, opened 2019-02-07
[interpreter] Return instance wrapper from instance()

binji:
Previously instances were stored by name as local variables. If we store
instances keyed by name, then we can issue the same commands on a
Worker. This will ensure that the Worker's state matches the state of
the main thread.

### DIFF
--- a/interpreter/script/js.ml
+++ b/interpreter/script/js.ml
@@ -28,15 +28,10 @@ let handler = {
     return (prop in target) ?  target[prop] : {};
   }
 };
-let instances = new Map();
 let registry = new Proxy({spectest}, handler);
 
-function getInstance(name) {
-  return instances.get(name);
-}
-
-function register(name, instanceName) {
-  registry[name] = getInstance(instanceName).exports;
+function register(name, instance) {
+  registry[name] = instance.exports;
 }
 
 function module(bytes, valid = true) {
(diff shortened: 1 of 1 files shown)


interpreter/script/js.ml
```diff
@@ -52,21 +57,36 @@ function module(bytes, valid = true) {
   return new WebAssembly.Module(buffer);
 }
 
-function instance(bytes, imports = registry) {
+function anonInstance(bytes, imports = registry) {
   return new WebAssembly.Instance(module(bytes), imports);
 }
 
-function call(instance, name, args) {
+function instance(names, bytes, imports = registry) {
```
rossberg: It would feel cleaner to me if the harness did not to mix creation of the instance with storing it under a name. Simply introduce a separate function, say, `setInstance`, and leave all code as before except replacing the generation of `let` with the generation of a call to this function. Similarly, move calls to `getInstance` to the generated code instead of coupling them with the call/get actions. Such a separation makes it easier to customise the harness, I think.

In addition, you could simply make `instances` an object, then you don't need the get/setInstance helpers but just use `instances["name"]` where it used `name` before.

binji:
> How will workers replay commands, and why does it matter how/where they find their instances locally?

The way I was thinking it, the `instance` function would do something like this:

```js
function instance(name, bytes) {
  const module = new Module(bytes);
  const message = {kind: 'instance', name, module};
  for (let worker in workers) { worker.postMessage(message); }
  const instance = new Instance(module);
  instances.set(name, instance);
  return instance;
}
```

Then the worker would receive the message:

```js
onmessage = function(event) {
  switch (event.data.kind) {
    case 'instance':
      const instance = new Instance(event.data.module);
      instances.set(name, instance);
      break;
    case 'call': ...
  }
}
```

interpreter/script/js.ml
binji: It's coupled so the name of the instance is associated with the module bytes / module creation, since I want to send the module associated with the name, not the instance.

Another alternative is to store the module on the instance object so it can be accessed from `setInstance`, I suppose.

rossberg:
Oh, I see! That makes sense.

How about having `instance` return a pair of module and instance?

binji:
I think that will work, will try.

binji:
I realized that we don't really need to send the name to the `Worker`, if we return an instance object wrapper, we can store a unique identifier which can later be sent. We could also store these values directly on the `WebAssembly.Instance` object, but that seems uglier.

PR #128, opened 2019-02-13
Update rmw instruction syntax to match wabt

alexcrichton:
Changed in WebAssembly/wabt#985


binji:
Good point, though this should be updated in the interpreter and spec too. We can land this and I'll update those afterward.

alexcrichton:
Oops, thanks for the follow-ups!

PR #131, opened 2019-02-14
[interpreter] Implement trivial wait/notify cases

binji:
`*.atomic.wait` only waits if the loaded value is equal to the expected
value. So we can test some simple cases without threads by ensuring that
the loaded value doesn't match.

Similarly `atomic.notify` will do nothing if the count is 0.

### DIFF
--- a/interpreter/exec/eval.ml
+++ b/interpreter/exec/eval.ml
@@ -276,6 +276,25 @@ let rec step (c : config) : config =
           v1 :: vs', []
         with exn -> vs', [Trapping (memory_error e.at exn) @@ e.at]);
 
+      | AtomicWait {offset; ty; sz; _}, I64 timeout :: ve :: I32 i :: vs' ->
+        let mem = memory frame.inst (0l @@ e.at) in
+        let addr = I64_convert.extend_i32_u i in
+        (try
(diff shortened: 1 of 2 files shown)


interpreter/exec/eval.ml
```diff
@@ -276,6 +276,25 @@ let rec step (c : config) : config =
...
         with exn -> vs', [Trapping (memory_error e.at exn) @@ e.at]);
 
+      | AtomicWait {offset; ty; sz; _}, I64 timeout :: ve :: I32 i :: vs' ->
+        let mem = memory frame.inst (0l @@ e.at) in
+        let addr = I64_convert.extend_i32_u i in
+        (try
+          assert (sz = None);
+          check_align addr ty sz e.at;
+          let v = Memory.load_value mem addr offset ty
+          in if v = ve then
```
rossberg: Nit: `in` on previous line

interpreter/exec/eval.ml
```diff
@@ -276,6 +276,25 @@ let rec step (c : config) : config =
...
+            assert false  (* Unimplemented *)
```
rossberg: Change comment to `TODO`, similarly below

PR #134, opened 2019-02-20
Improve tests for evaluation order.

### LINKED ISSUE #133: [test] js-api tests getter on Boolean
This is already tracked in v8 but since it's duplicated here I'm posting it too.

According to https://tc39.github.io/ecma262/#sec-toboolean, https://github.com/WebAssembly/threads/blob/0b80037ebc7825db0e67c2433f556dd8d07118a2/test/js-api/memory/constructor.any.js#L133 won't be called.

Ms2ger:
Fixes #133.

### DIFF
--- a/test/js-api/memory/constructor.any.js
+++ b/test/js-api/memory/constructor.any.js
@@ -125,14 +125,44 @@ test(() => {
         },
       };
     },
+  });
+
+  assert_array_equals(order, [
+    "initial",
+    "initial valueOf",
+    "maximum",
+    "maximum valueOf",
+  ]);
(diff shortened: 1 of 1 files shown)


xtuc:
I was about to add the test:
```js
// use this ordering test when using boolean types
test(() => {
  const order = [];

  const mem = new WebAssembly.Memory(new Proxy({}, {
    get(o, x) {
      order.push(x);
      return 0;
    },
  }));

  assert_array_equals(order, ["initial", "maximum", "shared"]);
}, "Order of evaluation for descriptor with shared");
```

Issue #140, opened 2019-06-16
atomic.fence validation and binary encoding


conrad-watt:
After the last CG, we agreed to add atomic.fence.

Validation-wise, the instruction has type ([] -> []). Unlike other atomic.* instructions, it is **_not_** a validation error for the instruction to occur in a module with a non-shared memory. This is because the instruction fences _all_ memories in the store, not just that of the current module.

Encoding-wise, I hope someone with better intuition about the binary format can suggest something.

tlively:
I don't quite understand why fence is allowed to be used with non-shared memories but other atomic instructions are not. Can you elaborate on that a bit?

sunfishcode:
The [C](https://en.cppreference.com/w/c/atomic/atomic_thread_fence), [C++](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence) and [Rust](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html) fence constructs have an "order" parameter. The current wasm threads proposal only supports SC, but other orderings may be added in the future. Would it make sense to include an immediate field in the fence encoding, to allow for other orderings to be specified in the future?

rossberg:
@tlively, a fence is not tied to any memory, but commits *all* accesses (which may involve multiple memories or even other state than memories in the future) that have occurred in the thread (which may include accesses perfomed in other modules on the same thread). So it isn't meaningful to require a specific memory to be present locally.

conrad-watt:
> The [C](https://en.cppreference.com/w/c/atomic/atomic_thread_fence), [C++](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence) and [Rust](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html) fence constructs have an "order" parameter. The current wasm threads proposal only supports SC, but other orderings may be added in the future. Would it make sense to include an immediate field in the fence encoding, to allow for other orderings to be specified in the future?

I think this makes sense. The same point holds for other atomic operations - any `order` encoding for fence should try to be consistent with what they are/will be doing. I can't see anything explicit, but is the `memarg` part of the existing atomics' binary encoding intended to be used this way in the future? (https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#spec-changes)

binji:
> I can't see anything explicit, but is the memarg part of the existing atomics' binary encoding intended to be used this way in the future? 

[...]

PR #141, opened 2019-06-19
Add fence instruction and encoding to overview

conrad-watt:
Following https://github.com/WebAssembly/threads/issues/140

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -439,9 +439,9 @@ For the web embedding, `atomic.notify` is equivalent in behavior to executing th
 
 ## Fence operator
 
-The fence operator, `atomic.fence`, takes no operands, and returns nothing. It is intended to preserve the semantics of the [fence operators of higher-level languages](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence).
+The fence operator, `atomic.fence`, takes no operands, and returns nothing. It is intended to preserve the synchronization guarantees of the [fence operators of higher-level languages](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence).
 
-Unlike other atomic operators, `atomic.fence` does not target a particular linear memory, and it is therefore *not* a validation error for this operator to be used in a module that does not declare a shared memory.
+Unlike other atomic operators, `atomic.fence` does not target a particular linear memory. It may occur in modules which declare no memory, or a non-shared memory, without causing a validation error.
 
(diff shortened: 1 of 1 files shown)


proposals/threads/Overview.md
```diff
@@ -590,6 +598,8 @@ instr ::= ...
         | 0xFE 0x01 m:memarg32  =>  i32.atomic.wait m
         | 0xFE 0x02 m:memarg64  =>  i64.atomic.wait m
 
+        | 0xFE 0x0F 0x00        =>  atomic.fence
```
lars-t-hansen: There's really no reason why this should be 0x0F and not 0x03.  It's already part of the 0xFE 0x0_ group, and we might as well pack each group densely.

conrad-watt: I was thinking that if we added something like a fork instruction, it would be nice to still have fence be closest to the load/store definitions. I agree it's not very strong reason though.

I've changed it to 0x03 now.

lars-t-hansen: Reasonable people can disagree about this, but personally I think it doesn't gain us anything to try to group the encodings of operations that have related functionality.  All of us are going to implement dispatch by lookup tables anyway, and over time there will be enough encoding exceptions that any system will come to seem non-coherent.

proposals/threads/Overview.md
```diff
@@ -437,6 +437,12 @@ For the web embedding, `atomic.notify` is equivalent in behavior to executing th
 1. Let `result` be [`Atomics.notify`][](`int32array`, `address`, `fcount`).
 1. Return `result` converted to an `i32`.
 
+## Fence operator
+
+The fence operator, `atomic.fence`, takes no operands, and returns nothing. It is intended to preserve the semantics of the [fence operators of higher-level languages](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence).
+
+Unlike other atomic operators, `atomic.fence` does not target a particular linear memory, and it is therefore *not* a validation error for this operator to be used in a module that does not declare a shared memory.
```
binji: nit: I assume it's also OK to be used in a module that declares no memory at all.

conrad-watt: yes; clarified this

conrad-watt:
Is there anything I can do to move this forward? I've addressed previous comments.

lars-t-hansen:
@binji, can you review too?

PR #150, opened 2020-02-28
Fix interpreter/tests to allow unshared memory

binji:


interpreter/valid/valid.ml
```diff
@@ -137,10 +137,7 @@ let type_cvtop at = function
 
 (* Expressions *)
 
-(* TODO: Remove sharability requirements *)
-let check_memop (c : context) (memop : 'a memop) (sh: sharability option)
-    get_sz at =
-  let MemoryType (_, shared) = memory c (0l @@ at) in
```
rossberg: This seems to remove the entire check that memory 0 exists. Isn't that still needed? I'm surprised that this breaks no tests.

binji: Whoops! Yeah, it breaks the memory.wast test. Fixed

binji:
The wait instructions have only been implemented for trivial cases. The other instructions have been implemented somewhat, but there is no actual concurrency happening.

PR #154, opened 2020-03-25
Throw TypeError for shared memory w/o maximum

### LINKED ISSUE #153: WebAssembly.Memory constructor exception
If shared is true and maximum isn't given, the Bikeshed says to throw a RangeError, but #24 and implementations aligned on TypeError. Probably makes the most sense to fix the specification here.

binji:
Fixes #153.


binji:
Oops, you're right :) I'll fix.

PR #158, opened 2020-04-09
Add a note for WebAssembly.Memory.prototype.grow return value

dtig:
The current text specifies the return value to be the current size of memory in pages (before resizing), but this is somewhat ambiguous and can lead to implementation specific behavior. For example, engines can decide to return the value at the time the grow method was called, or at the time the memory was updated, and these can be different values depending on the strictness of the implementation itself, and if there are multiple grow calls that occur at the same time. 

Add a note to clarify that the return value should essentially be an atomic RMW to the length of the underlying buffer. The implication of this is that more than one grow call cannot return the same value.

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -514,9 +514,12 @@ thrown.
 
 Return `ret` as a Number value.
 
-Note: When [`IsSharedArrayBuffer`][](`M.[[BufferObject]]`) is true, the value `ret`
-should be the result of an atomic read-modify-write of the new size to the
-internal [\[\[ArrayBufferByteLength\]\]][] slot. 
+Note: When [`IsSharedArrayBuffer`][](`M.[[BufferObject]]`) is true, the return
+value should be the result of an atomic read-modify-write of the new size
+to the internal [\[\[ArrayBufferByteLength\]\]][] slot. The `ret` value will be
+the value in pages read from the internal [\[\[ArrayBufferByteLength\]\]][] slot 
+before the modification to the resized size, which will be the current size of
+the memory.
 
 ### `WebAssembly.Memory.prototype.buffer`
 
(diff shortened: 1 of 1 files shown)


proposals/threads/Overview.md
```diff
@@ -514,6 +514,10 @@ thrown.
 
 Return `ret` as a Number value.
 
+Note: When [`IsSharedArrayBuffer`][](`M.[[BufferObject]]`) is true, the value `ret`
+should be the result of an atomic read-modify-write of the new size to the
```
binji, line 518: The only issue here is that a RMW doesn't always return the old value, IIRC (it does in wasm and JS but not in other implementations). Maybe worth clarifying here.

dtig: Done

lars-t-hansen:
This language is not completely satisfactory, because (a) the Memory.grow algorithm (which is referenced here, though the link is 404) does not traffic in any `[[ArrayBufferByteLength]]` slot - that a JS-ism, and (b) the RMW needs a consistency attribute to be fully specified and (c) as @binji alluded to, Memory.grow must read the old length, compute on it to validate it, and then store the new value, and saying that it must be an "atomic rmw operation" is somewhat misleading - really it would have to be an atomic-ish read, followed later by an atomic compare-exchange, but even then it begs the question of how other state is updated.

It might be less ambiguous and less implementation-oriented to say that there is a per-memory critical section around the two steps of this algorithm that read the current length and then call the Memory.grow, thus including all the steps of Memory.grow.  Then implementations are free to hoist activities out of the critical section so long as they do not change the semantics.

(I've assumed in the past that people agree on the meaning of "critical section" and discovered that it ain't necessarily so, but even so I think this is less problematic.)

dtig:
Thanks @lars-t-hansen for the feedback, all three points mentioned above make sense for (a) specifically though, as this document mostly addresses the JS API changes, I addressed the JS terminology here, and figured that when memory.ml is updated to reflect changes specific to this proposal, that would be updated as well - but I can see why this is ambiguous. I'll follow up with a PR to try and make this less ambiguous. 

Fix for broken links updated here - https://github.com/WebAssembly/threads/pull/159

PR #167, opened 2020-10-26
Merge upstream

rossberg:
On the way, fixed some new tests in binary.wast: they were supposed to check malformed table limits but erroneously defined a memory section, so failed for the wrong reason. Also removed a few tests that each contained 2 competing bugs but were otherwise redundant.


binji (approved):
rubber stamp lgtm. But can you link to the commits where you made other changes? There's a lot here to scrub through

rossberg:
Sorry about that. All new changes are in the merge commit [739a2e7](https://github.com/WebAssembly/threads/pull/167/commits/739a2e7923efec5589fb6e9f05e3e002c3f2ff55) itself, but I'm not sure how you can see its diff in isolation on GH. And even looking at that commit offline now with `git show` or `git diff-tree` the result includes huge hunks with tons of spurious lines from parents, such as the entire gen-index-instructions file that I had only extended.

Is there a better way of doing merges? I usually try not to commit a merge before all conflicts are resolved and the whole thing builds again. But it seems like that defeats reviewing. A quick Web search didn't provide a good answer.

binji:
Yeah, I'm not sure there is a better way. It looks like `git show <merge commit>` does the right thing from the git CLI (I believe it uses `--cc` implicitly?), but this doesn't seem to be the case with GitHub.

rossberg:
Yes, `git show` is essentially `git diff-tree --cc` IIUC. But that's a three-way diff, in this case 2.5K lines large, because a small edit to a big change makes it includes the entire thing as a big hunk.

PR #169, opened 2020-10-26
Update runtime; tweaks

rossberg:
Various tweaks, updates, and todos to get the ball rolling again. In particular, split store into shared and local parts and switch to symbolic addresses.


document/core/exec/runtime.rst
```diff
@@ -277,24 +315,26 @@ It is an invariant of the semantics that the length of the element vector never
 Memory Instances
 ~~~~~~~~~~~~~~~~
 
+.. todo:: We used to update the memory type when a memory is grown. This does not work for shared memories. In fact, the "current" type of a shared memory is nondeterministic. We need to model that during instantiation somehow.
```
conrad-watt, line 318: I had assumed that instantiation would proceed as though executing `memory.size` on each imported memory (similar to the way global initialisers are evaluated), but I realise now that this hasn't really been discussed.

IIUC, the memory type is only introspected during instantiation, so for instruction validation and the soundness statement, it would be OK for the memory type to no longer contain a `min` size?

document/core/exec/runtime.rst
```diff
@@ -277,24 +315,26 @@ It is an invariant of the semantics that the length of the element vector never
...
-A memory instance records an optional maximum size for a memory, if one was specified at the definition site of the memory, and a flag indicating whether the memory can be :ref:`shared <syntax-shared>` across threads.
+It records its original :ref:`memory type <syntax-memtype>`
+and takes one of two different forms depending on whether that type is :ref:`shared <syntax-shared>` or not.
 
 .. math::
    \begin{array}{llll}
    \production{(memory instance)} & \meminst &::=&
-     \{ \MIMAX~\u32^?, \MISHARE~\UNSHARED, \MIDATA~\vec(\byte) \} \\ &&|&
-     \{ \MIMAX~\u32^?, \MISHARE~\SHARED \} \\
+     \{ \MITYPE~\memtype, \MIDATA~\vec(\byte) \} \\ &&|&
```
conrad-watt, line 327: syntactically there's nothing here stopping memtype from being `shared`, even if there's a byte vector, right?

document/core/exec/runtime.rst
```diff
@@ -702,38 +696,115 @@ Its form and meaning is outside the scope of this specification.
...
-     \act^\ast~\AT~\time \\
+   \config \stepto^{\evt^\ast} \config
+
+defined by inductive rewrite rules on global configurations.
+
+The following structural rule for global reduction delegates to local reduction for single thread execution:
+
+.. math::
+   \begin{array}{@{}c@{}}
+   \XX{SS}; P_1^\ast~(S; \instr^\ast \AT h)~P_2^\ast
```
conrad-watt, line 768: I might be misreading this rule (I'm having trouble getting the markup building to work) but why is the empty frame `F_\epsilon` needed? I would have thought that each thread would have its own frame (as defined [here](https://webassembly.github.io/spec/core/exec/runtime.html?highlight=thread#configurations)) instead of just being local store + instructions.

document/core/appendix/properties.rst
```diff
@@ -745,10 +745,23 @@ In other words, every thread in a valid configuration either runs forever, traps
 Consequently, given a :ref:`valid store <valid-store>`, no computation defined by :ref:`instantiation <exec-instantiation>` or :ref:`invocation <exec-invocation>` of a valid module can "crash" or otherwise (mis)behave in ways not covered by the :ref:`execution <exec>` semantics given in this specification.
 
 
+.. index:: relaxed memory, ! sequential consistency
+
+Sequential Consistency of Data-Race-Free Programs
```
conrad-watt, line 750: If you remember the corrigendum I sent you (still going through the ACM process), we have to be a little apologetic about SC-DRF - the property we inherit from JS is weaker than the ideal one (which is technically false, because of a corner to do with thin-air).

I should handle writing something about this - basically a non-normative note _somewhere_ (appendix?) to the effect that because of known limitations in our ability to formalise relaxed memory, the model has gaps in it which are subjects of active research.

conrad-watt (commented):
Had an initial look, still trying to get the document to build locally so I can properly see the markup.

The only actionable question I have is about the global reduction rule (see below). Other comments can probably be addressed later in drafting.

document/core/exec/runtime.rst
rossberg: Memory types are used for a number of purposes, and its notion of limits is furthermore shared with tables. I'd rather not introduce multiple different variations and intersections of these concepts. Rather, some form of abstraction to read the current "dynamic" type of a memory at instantiation time. That would also be what's used by the embedder API.

document/core/exec/runtime.rst
rossberg: Yes, it's just an invariant of the semantics (I actually had a statement about that initially, maybe I should reinstate it for clarity).

document/core/exec/runtime.rst
rossberg: Well, then you would just permanently keep all the empty frames in the global configuration (and formulate well-formedness rules for them), and you'd need to create them during thread creation.

The main reason I did it like that in the original formulation was that there was only one level of reduction relation, so no natural place to inject the empty frame as you go. But now that we have that, it seems less attractive to keep that artefact around. The only appeal would be that thread and local configuration get unified, but is that valuable? I don't know, though I could be convinced otherwise.

Btw, I don't know if you noticed, but despite the split store, the rule here is currently simpler than the one in our paper:

* It does not allow any allocation, i.e., the global store remains unchanged. At least for now, the only operation that allocates is instantiate, which is simpler to model with a global reduction rule.

* It also doesn't pass the prior global store to the local reduction, since the operations where it would be needed (mutation of a shared table or global) are not currently supported.

So this completely decouples local reduction from the global store for now, which is nice. Even if we add allocation forms, it might actually be simpler to do so through directly through global reductions. (Typed updates will remain an annoyance, though.)

rossberg (commented):
I can't build myself right now, our Sphinx hacks bit-rot too quickly. :( The problems should be fixed with the merge in #167, though.

document/core/exec/runtime.rst
conrad-watt: My thinking was that unifying threads and local configurations would mean they wouldn't need separate validity rules in the soundness statement. It's a very small thing either way.

Rest of your comment looks good!

document/core/appendix/properties.rst
conrad-watt: agree

document/core/exec/runtime.rst
conrad-watt: to be bikeshedded when instantiation is specified

conrad-watt (approved):
LGTM then if the document can't be built right now

WDYT about my point regarding the empty frame in the global reduction rule?

rossberg:
> WDYT about my point regarding the empty frame in the global reduction rule?

We can always change that later, it doesn't affect much else, I think. Right now my slight preference is to keep it this way, because conceptually, threads do not really have a frame -- that's an attribute of function activations, and a local configuration nicely corresponds to the current activation. Inserting the empty frame in this rule encapsulates this auxiliary device to bridge the gap within a single rule.

I'll wait with landing until #167 is merged and I have rebased this PR.

rossberg:
@conrad-watt, I merged upstream and fixed the latex, should build again. Also adjusted rules to the changed meminst def that I overlooked before. PTAL at the last commit.

document/core/exec/modules.rst
```diff
@@ -325,7 +327,7 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
 
 3. Let :math:`a` be the first free :ref:`memory address <syntax-memaddr>` in :math:`S`.
 
-4. Let :math:`\meminst` be the :ref:`memory instance <syntax-meminst>` :math:`\{ \MIDATA~(\hex{00})^{n \cdot 64\,\F{Ki}}, \MIMAX~m^?, \MISHARE~s\}` that contains :math:`n` pages of zeroed :ref:`bytes <syntax-byte>`.
+4. Let :math:`\meminst` be the :ref:`memory instance <syntax-meminst>` :math:`\{ \MITYPE~\memtype, \MIDATA~(\hex{00})^{n \cdot 64\,\F{Ki}} \}` that contains :math:`n` pages of zeroed :ref:`bytes <syntax-byte>`.
```
conrad-watt, line 330: Do you think it's worth changing this now so that the `\MIDATA` component is only present if the memory type is `unshared`, or is that part of the above todo?

document/core/exec/modules.rst
```diff
@@ -337,7 +339,7 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
    \mbox{where:} \hfill \\
    \memtype &=& \{\LMIN~n, \LMAX~m^?\}~s \\
    \memaddr &=& |S.\SMEMS| \\
-   \meminst &=& \{ \MIDATA~(\hex{00})^{n \cdot 64\,\F{Ki}}, \MIMAX~m^?, \MISHARE~s\} \\
+   \meminst &=& \{ \MITYPE~\memtype, \MIDATA~(\hex{00})^{n \cdot 64\,\F{Ki}} \} \\
```
conrad-watt, line 342: same comment as above

document/core/exec/relaxed.rst
```diff
@@ -1,8 +1,30 @@
...
+
+.. todo:: define, explain
+
+
+.. _relaxed-valid:
+
+Validity
+~~~~~~~~
+
+.. todo:: rename to avoid confusion with validation
```
conrad-watt, line 26: We could rename "validity" to "consistency". The former is ARM model terminology, the later is C++ model terminology.

document/core/exec/modules.rst
```diff
@@ -59,15 +59,17 @@ The following auxiliary typing rules specify this typing relation relative to a
...
 
-* Then :math:`\EVMEM~a` is valid with :ref:`external type <syntax-externtype>` :math:`\ETMEM~(\{\LMIN~n, \LMAX~m^?\}~s)`.
+* The store entry :math:`S.\SMEMS[a]` must be a :ref:`memory instance <syntax-meminst>` :math:`\{\MITYPE~\{\LMIN~n, \LMAX~m^?\}~s, \MIDATA~b^{k\cdot64\,\F{Ki}}\}`.
+
+* Then :math:`\EVMEM~a` is valid with :ref:`external type <syntax-externtype>` :math:`\ETMEM~\{\LMIN~k, \LMAX~m^?\}~s`.
 
 .. math::
    \frac{
-     S.\SMEMS[a] = \{ \MIDATA~b^{n\cdot64\,\F{Ki}}, \MIMAX~m^?, \MISHARE~s\}
+     S.\SMEMS[a] = \{ \MITYPE~\{\LMIN~n, \LMAX~m^?\}~s, \MIDATA~b^{k\cdot64\,\F{Ki}} \}
```
conrad-watt, line 70: What's the reasoning for this change? It's a little strange that the memory instance now has a "minimum" component which is never used, and it makes the definition of external typing more surprising. The change is also not currently mirrored in the table instance, although it could be.

It does make the soundness proof somewhat neater (give the related change to memory instance validity). Is that why?

conrad-watt: I just realised the change to the syntax of the memory instance was made before and I missed it - I should have asked this question them

document/core/exec/modules.rst
rossberg: Yes, I think adapting this requires a larger refactoring. For now I just adjusted the syntax to be consistent (and to avoid Latex errors due to missing defs).

document/core/exec/relaxed.rst
rossberg: Consistency sounds good to me! I went ahead and renamed it.

document/core/util/macros.def
```diff
@@ -1068,17 +1074,19 @@
 
 .. Relaxed Memory Model Judgements
 
-.. |valid| mathdef:: \mathrel{\mbox{valid}}
-.. |validwith| mathdef:: \mathrel{\mbox{valid-with}}
+.. todo:: include xrefs
```
conrad-watt, line 1077: Is this related to the CI builds failing?

conrad-watt (approved):
LGTM!

document/core/util/macros.def
rossberg: Probably. I moved the todo elsewhere.

Issue #163, opened 2020-08-19
Requirements for phase 3


tlively:
This repo has not had a lot of activity lately, so I'm wondering if this proposal is ready to move to phase 3.

The phase 3 entry requirements are:

 1. Test suite has been updated to cover the feature in its forked repo.

 2. The test suite should run against some implementation, though it need not be the reference interpreter.

We certainly have a working implementation, and it looks like a test suite has been added. Is there anything else we previously decided to wait on before moving to phase 3?

lars-t-hansen:
I thought the process required the spec text but it does not, so we don't have to wait on the memory model for phase 3, for one thing.

(FF, like Chrome, has shipped this feature.)

binji:
Good point, I had assumed we were waiting on some additional spec work to advance, but this looks like it can move to phase 3. The testsuite is on concern, however -- we have surface-level tests of the new instructions, but we'll really want something more comprehensive. Maybe it's sufficient for phase 3 if we can move some of Chrome's and FF's JS tests into this repo.

Constellation:
Note that JSC implemented this behind the flag.
https://trac.webkit.org/changeset/270208/webkit

dtig:
Looks like this discussion has stalled, was this discussed in a meeting with additional follow up? If not, I'm wondering if in addition to the existing spec tests, would [WPT](https://web-platform-tests.org/) be a good place to start to make sure that everything works as it should with workers? 

My concern with JS Shell tests is that while these will work, they are not reflective of real world cases i.e. at least in d8, the worker API is somewhat barebones and depending on test cases it may be possible to get different results when running in the shell vs. the browser, and while we can fix the API, that seems to me to be the wrong thing to fix in this case. I'm probably missing quite a bit of context w.r.t. testing, but I'm wondering what we can do in the near term to hit phase 3 requirements. Concretely I'd like to propose that we decouple comprehensive testing and figuring out which high level test predicate should be used - i.e. use WPT in the near term for comprehensive testing for phase 3, and work on the larger problem of implementing/using the API linked above, or a different strategy for higher level testing separately.

tlively:
> Looks like this discussion has stalled, was this discussed in a meeting with additional follow up?

It was not.

rossberg:
The test suite requirement should be read as tests integrated into the wast-based spec test suite. A JS-specific test suite is of little use to non-JS engines.

[...]

PR #179, opened 2021-07-26
[interpreter] Threading

rossberg:
This adds threading support to the interpreter:

* Add threads to evaluator and script runner, with randomised execution.
* Add `thread` and `wait` script commands for spawning and joining threads.
* Add `either` result pattern for testing non-deterministic results.
* Extend JS harness with new functions `thread` and `wait` (not yet fleshed out).
* Add a trivial test.
* Added command line flags for setting non-determinism random seed.

This includes scripting functionality as discussed in #163. However, I decided to make sharing more explicit with `(sharing ...)` annotations in a thread definition, because it's more readable, avoids funny errors, and it's probably gonna be easier to implement in the JS converter that way. See test for how it's used.

Sorry for the large PR, this required more refactoring than I anticipated. In particular, the whole script runner had to effectively be transformed from big step to small step. Similarly, Eval.invoke and Eval.init needed to become small steps.

### DIFF
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -397,7 +397,7 @@ assertion:
 
 result:
   ( <val_type>.const <numpat> )              ;; numeric result
-  ( oneof result* )                          ;; non-deterministic result
+  ( either result+ )                         ;; non-deterministic result
 
 numpat:
   <value>                                    ;; literal result
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -374,6 +374,8 @@ binscript: <cmd>*
 cmd:
   <module>                                   ;; define, validate, and initialize module
   ( register <string> <name>? )              ;; register module for imports
+  ( thread <name>? cmd* )                    ;; spawn thread
+  ( wait <name> )                            ;; join thread
   <action>                                   ;; perform action and print results
   <assertion>                                ;; assert result of an action
 
@@ -394,7 +396,8 @@ assertion:
   ( assert_trap <module> <failure> )         ;; assert module traps on instantiation
 
 result:
-  ( <val_type>.const <numpat> )
+  ( <val_type>.const <numpat> )              ;; numeric result
+  ( oneof result* )                          ;; non-deterministic result
 
 numpat:
   <value>                                    ;; literal result
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -374,10 +374,13 @@ binscript: <cmd>*
 cmd:
   <module>                                   ;; define, validate, and initialize module
   ( register <string> <name>? )              ;; register module for imports
-  ( thread <name>? cmd* )                    ;; spawn thread
+  ( thread <name>? <shared>* cmd* )          ;; spawn thread
   ( wait <name> )                            ;; join thread
   <action>                                   ;; perform action and print results
   <assertion>                                ;; assert result of an action
+
+shared:
+  ( shared ( module <name> ) )
 
 module:
   ( module <name>? binary <string>* )        ;; module in binary format (may be malformed)
--- a/interpreter/main/main.ml
+++ b/interpreter/main/main.ml
@@ -42,13 +42,13 @@ let () =
     configure ();
     Arg.parse argspec
       (fun file -> add_arg ("(input " ^ quote file ^ ")")) usage;
-    let config = Run.config () in
-    List.iter (fun arg -> if not (Run.run_string config arg) then exit 1) !args;
+    let context = Run.context () in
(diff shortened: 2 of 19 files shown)


SuspendedPhan:
Sorry for the noise, but just wanted to say seriously thanks for working on this! golang said they would implement threads once this moves to stage 3 :D

-- excited adopter of wasm

interpreter/script/run.ml
```diff
@@ -248,18 +248,22 @@ let string_of_nan = function
   | CanonicalNan -> "nan:canonical"
   | ArithmeticNan -> "nan:arithmetic"
 
-let type_of_result r =
-  match r with
+let rec type_of_result r =
+  match r.it with
   | LitResult v -> Values.type_of v.it
   | NanResult n -> Values.type_of n.it
+  | EitherResult rs -> type_of_result (List.hd rs)
```
conrad-watt, line 255: There appears to be some jankiness here - for example, if I change the `either` line in `thread.wast` to
```
  (assert_return (invoke "run") (either (i32.const 0) (i64.const 42)))
```

The test will randomly either pass, or fail with the message
```
> Result: 42 : i32
> Expect: (0 | 42) : i32
> -- Error: 
> ../test/core/thread.wast:24.3-24.71: assertion failure: wrong return values
```

Not sure what the right approach to polish this is... force all elements of `either` to have the same type?

interpreter/README.md
```diff
@@ -394,7 +399,8 @@ assertion:
   ( assert_trap <module> <failure> )         ;; assert module traps on instantiation
 
 result:
-  ( <val_type>.const <numpat> )
+  ( <val_type>.const <numpat> )              ;; numeric result
+  ( either result+ )                         ;; non-deterministic result
```
conrad-watt, line 403: This seems to have an unfortunate interaction with multi-value - because the either combinator is per-value, I can't express something like "either [0,0], or [1,42]". Instead I can only express [(either 0 or 1),(either 0 or 42)] which is less precise.

conrad-watt (changes requested):
Is it possible to lift the `either` combinator so that it applies to lists of values, rather than individual values (see comment on multi-value below)?

interpreter/main/main.ml
```diff
@@ -31,6 +31,8 @@ let argspec = Arg.align
   "-h", Arg.Clear Flags.harness, " exclude harness for JS conversion";
   "-d", Arg.Set Flags.dry, " dry, do not run program";
   "-t", Arg.Set Flags.trace, " trace execution";
+  "-r", Arg.Int Random.init, " set non-determinism random seed";
+  "-rr", Arg.Unit Random.self_init, " randomize non-determinism random seed";
```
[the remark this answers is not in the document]
rossberg: The docs aren't terribly clear on that, but I'm pretty sure there is a deterministic default seed when you don't set one explicitly. At least all my runs always produce the same sequence of random numbers. So I believe it should be fine as is.

interpreter/script/run.ml
rossberg: You are picky. :) You can get the same error without using `either`. The script level is dumb and does no type checking (which actually is a feature, since the interpreter can intentionally run unvalidated code with -u). What's a bit misleading here, though, is the random type annotation that gets printed. I extended that to check if the either types are consistent and print "?" otherwise. Though once we merge upstream, this will require computing a lub to be accurate, so not sure we'll really want to keep bothering...

interpreter/README.md
rossberg: Yeah, I considered lifting it, but there is no obvious syntax for writing the list of lists. In the end I decided that it's probably not worth the effort, since you could always formulate tests such that they encode the cases in a single value. We were doing that routinely before multi values.

interpreter/main/main.ml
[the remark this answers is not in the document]
conrad-watt: Ah you're right, OCaml itself seems to start with a hardcoded seed (https://github.com/lucasaiu/ocaml/blob/master/stdlib/random.ml#L149)

conrad-watt (approved):
LGTM

I see that `wait/notify` implementations are stubbed out - would these need refactorings to implement?

interpreter/README.md
conrad-watt: make sense

interpreter/script/run.ml
conrad-watt: I think I'm comfortable being imprecise with `?` (even when there's a LUB) rather than printing the wrong type - change makes sense!

rossberg:
> I see that `wait/notify` implementations are stubbed out - would these need refactorings to implement?

Good question, I don't know. I hope not.

PR #190, opened 2022-09-07
Update description of `memory.atomic.wait64`

tlively:
Now that JS `Atomics.wait` natively supports BigInt64Array,
`memory.atomic.wait64` can be described directly in terms of `Atomic.wait`.


conrad-watt (approved):
LGTM

Issue #144, opened 2019-10-17
Allowing atomic instructions with unshared memory


tlively:
It is currently specified that attempting to use atomic access operators on non-shared linear memory is a validation error. This validation requirement causes a lot of friction for toolchain developers and users.

For example, consider a user wishing to distribute a WebAssembly binary for a C++ library. Say that library uses C++11 atomics internally to achieve thread-safety, but is otherwise agnostic to whether it is used in a single threaded environment or a multithreaded environment. On native platforms a single build of the library can be used in both environments, but due to the validation rules for WebAssembly atomics, the library must be built twice for WebAssembly; once with atomics enabled and once without. Beyond causing extra work for developers, having to produce these separate builds is also a large hole in WebAssembly's portability story.

One solution to this would be for all applications to use shared memories whether they are multithreaded or not. But this is not practical because not all engines support threads and not all engines are expected to support threads in the future.

A better solution would be to relax the validation rules to allow atomic operations to be used with non-shared memories. The semantics for atomic loads, stores, rmw operations, and cmpxch operations are straightforward to extend to unshared memories. There are more possibilities that might make sense for wait and notify operations, but I suggest that waiting on an unshared memory always trap and notify on an unshared memory always return 0.

lars-t-hansen:
Making the load/store/rmw/cmpxchg ops more lenient seems fairly unproblematic (though see below) and would make simple cases "just work".

`wait` and `notify` I'm more skeptical about.  Anyone using `wait` and `notify` is basically implementing some sort of mutex, and it's not terribly plausible that that would work the way it's intended in unshared memory.  So then the argument would have to be that it's possible for the module to discover whether its memory is shared or not and hence whether to use the mutex or not.  So do we want a mechanism for that?  I can see the case for allowing these ops to verify in unshared memory but probably not the case for allowing them to execute without traps (even `notify`).

Another concern with the simple operations is that code written for shared memory may use techniques that are not directly related to `wait`/`notify`, such as spinlocks, that make little sense in unshared memory.  This again suggests a need for a mechanism to discover that shared memory is used.

It is possible that we could allow the simple ops to verify and execute with non-shared memory but that `wait`/`notify` won't even verify w/o shared memory and therefore must be used under control of feature detection, which then must know about shared memory.

[...]

PR #191, opened 2022-09-12
Soften language for `notify` on unshared memories

tlively:
Since #144, we've specified that `memory.atomic.notify` is allowed to be executed on unshared memories, but always returns 0 since it is impossible for there to be anything waiting on an unshared memory since all variations of the wait instructions in Wasm and JS trap or throw exceptions on unshared memories. However, I realized recently that there is no technical reason why [`Atomics.waitAsync`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/waitAsync) shouldn't be able to wait on an unshared memory. Because `Atomics.waitAsync` returns a promise rather than blocking, it can in principle be used to synchronize between multiple asynchronous tasks in flight concurrently in the same JS context, and now that [JSPI](https://github.com/WebAssembly/js-promise-integration/blob/main/proposals/js-promise-integration/Overview.md) is becoming more real, this actually might be a useful thing to do. In particular `Atomics.waitAsync` could be used to build suspending versions of mutexes and condition variables to synchronize between threads running as different asynchronous tasks within a single JS context.

Actually realizing this vision of asynchronous threading with JSPI on top of `Atomics.waitAsync` requires changes on the JS side to allow `Atomics.waitAsync` and `Atomics.notify` to be used on unshared ArrayBuffers, but in the meantime we can prepare for that change on the Wasm side by softening the language stating that `memory.atomic.notify` unconditionally returns 0 for unshared memories.

cc @syg

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -416,9 +416,10 @@ Similarly, `memory.atomic.wait64` is equivalent in behavior to executing the fol
 The notify operator takes two operands: an address operand and a count as an
 unsigned `i32`. The operation will notify as many waiters as are waiting on the
 same effective address, up to the maximum as specified by `count`. The operator
-returns the number of waiters that were woken as an unsigned `i32`. Note that if
-the notify operator is used with an unshared linear memory, the number of
-waiters will always be zero.
+returns the number of waiters that were woken as an unsigned `i32`. Note that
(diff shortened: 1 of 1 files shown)


conrad-watt (approved):
Makes sense, thanks!

conrad-watt:
(I'd also be interested in @syg's take on how likely this is on the JS side)

PR #192, opened 2022-09-29
[interpreter] implement atomic.fence, add wait/notify edge case tests

conrad-watt:


interpreter/exec/eval.ml
```diff
@@ -132,6 +132,10 @@ let check_align addr ty sz at =
   if not (Memory.is_aligned addr ty sz) then
     Trap.error at "unaligned atomic memory access"
 
+let check_shared mem at =
+  if (shared_memory_type (Memory.type_of mem)) != Shared then
```
rossberg, line 136:
```suggestion
  if shared_memory_type (Memory.type_of mem) <> Shared then
```
(`!=` is physical equality and almost never what you want for immutable values.)

interpreter/exec/eval.ml
```diff
@@ -319,18 +323,28 @@ let rec step_thread (t : thread) : thread =
...
           else
             I32 1l :: vs', []  (* Not equal *)
         with exn -> vs', [Trapping (memory_error e.at exn) @@ e.at])
 
-      | MemoryAtomicNotify x, I32 count :: I32 i :: vs' ->
+      | MemoryAtomicNotify {offset; ty; sz; _}, I32 count :: I32 i :: vs' ->
+        let mem = memory frame.inst (0l @@ e.at) in
+        let addr = I64_convert.extend_i32_u i in
+        (try
+          check_align addr ty sz e.at;
```
rossberg, line 338: Shouldn't this symmetrically require `shared` as well?

conrad-watt: Currently we allow `notify` on unshared memory. It's only `wait` that's disallowed.

interpreter/exec/eval.ml
conrad-watt: thanks!

conrad-watt:
@rossberg friendly ping

PR #194, opened 2022-10-19
[interpreter] implement atomic.wait and atomic.notify

conrad-watt:
also split out relevant tests into a separate file

### DIFF
--- a/interpreter/exec/eval.ml
+++ b/interpreter/exec/eval.ml
@@ -34,6 +34,8 @@ let numeric_error at = function
        string_of_int i ^ ", got " ^ Types.string_of_value_type (type_of v))
   | exn -> raise exn
 
+(* Must be positive and non-zero *)
+let timeout_epsilon = 1000000L
 
 (* Administrative Expressions & Configurations *)
 
@@ -57,11 +59,19 @@ and admin_instr' =
   | Label of int32 * instr list * code
   | Frame of int32 * frame * code
 
+type suspend_signal =
+    No_signal
+    (* memory, cell index, target timestamp *)
+  | Wait_signal of memory_inst * Memory.address * float
+    (* memory, cell index, number of threads to wake *)
+  | Notify_signal of memory_inst * Memory.address * int32
+
 type thread =
 {
   frame : frame;
   code : code;
   budget : int;  (* to model stack overflow *)
+  suspend_signal : suspend_signal
 }
 
 type config = thread list
@@ -69,7 +79,7 @@ type thread_id = int
 type status = Running | Result of value list | Trap of exn
 
 let frame inst locals = {inst; locals}
-let thread inst vs es = {frame = frame inst []; code = vs, es; budget = 300}
+let thread inst vs es = {frame = frame inst []; code = vs, es; budget = 300; suspend_signal = No_signal}
 let empty_thread = thread empty_module_inst [] []
 let empty_config = []
 let spawn (c : config) = List.length c, c @ [empty_thread]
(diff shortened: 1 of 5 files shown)


rossberg:
I'm somewhat confused, where is the notion of signal coming from? Is that an encoding of the reaction rules from the paper?

I would have expected a more direct implementation with administrative instructions wait' and notify', as on paper. And then perhaps auxiliary functions like the following to implement context matching:
```
match_wait' : code -> (memory_instance * address * f64 * (code -> code)) option
match_notify' : code -> (memory_instance * address * i64 * (code -> code)) option
```
to look for them (where the returned functions are "plug" function generated on the way to replace the found instruction with new code).

Also, the step function as written seems to initiate all reactions eagerly in a meta-level loop, whereas each wait/notify match is a separate step in the paper rules.

conrad-watt:
We discussed before that we wanted to shift to an axiomatic specification for wait/notify - i.e. emitting actions without blocking. These "signals" are intended to reflect the actions relevant for wait/notify in that formulation. I guess I could rename the datatype here to "action" to make the correspondance clearer.

rossberg:
Ah okay, that probably was a long time ago, so I don't recall. :) What would the reduction rules look like? If these signals correspond to actions, why do they become part of the thread representation in the interpreter?

conrad-watt:
I think you're right - they shouldn't be part of the `thread` representation. I've just changed things up to introduce an explicit `suspend` administrative instruction, and tweaked the step function to now return a pair of `thread * action`, removing the `action` from the `thread` representation itself.

The reduction rules will look something like (roughly)

```
wait -> i32 1
with act = load... (not matching)

wait -> suspend
with act = load... (matching) + wait ...
```

```
suspend -> i32 0
with act = woken ...

suspend -> i32 2
with <some timeout handling>
```

```
notify -> i32 n
with act = wakes n...
```

and then in the axiomatic memory model, all the `wait`, `woken`, and `wakes` actions have to satisfy a queue specification.

interpreter/exec/eval.ml
```diff
@@ -441,23 +455,72 @@ let rec step_thread (t : thread) : thread =
...
+  if count = 0l then
+    c, 0l
+  else
+    match c with
+    | [] ->
+      c, 0l
+    | t :: ts ->
+      let t', count' = (match (try_wake_one t m addr) with | None -> t, 0l | Some t'' -> t'', 1l) in
+      let ts', count'' = wake ts m addr (Int32.sub count count') in
+      (t' :: ts', Int32.add count' count'')
```
rossberg, line 509:
```suggestion
      t' :: ts', Int32.add count' count''
```

interpreter/exec/eval.ml
```diff
@@ -441,23 +455,72 @@ let rec step_thread (t : thread) : thread =
...
+
+let rec try_unsuspend (c : code) (m : memory_inst) (addr : Memory.address) : code option =
+  let vs, es = c in
+  match es with
+  | {it = Label (n, es0, c'); at} :: es' ->
+    Lib.Option.map (fun c'' -> vs, {it = Label (n, es0, c''); at} :: es') (try_unsuspend c' m addr)
+  | {it = Frame (n, f, c'); at} :: es' ->
+    Lib.Option.map (fun c'' -> vs, {it = Frame (n, f, c''); at} :: es') (try_unsuspend c' m addr)
+  | {it = Suspend (m', addr', timeout); at} :: es' ->
+    if Memory.is_same_memory m m' && addr = addr' then
```
rossberg, line 489:
```suggestion
    if m == m' && addr = addr' then
```

interpreter/exec/eval.ml
```diff
@@ -441,23 +455,72 @@ let rec step_thread (t : thread) : thread =
...
+      Some (((I32 0l) :: vs), es')
```
rossberg, line 490:
```suggestion
      Some (I32 0l :: vs, es')
```

interpreter/exec/eval.ml
```diff
@@ -441,23 +455,72 @@ let rec step_thread (t : thread) : thread =
...
+    else
+      None
+  | _ ->
+    None
+
+let try_wake_one (t : thread) (m : memory_inst) (addr : Memory.address) : thread option =
```
rossberg, line 496: This function seems a bit redundant, can't we inline it?

interpreter/exec/eval.ml
```diff
@@ -441,23 +455,72 @@ let rec step_thread (t : thread) : thread =
...
       )
-  in {t with code = vs', es' @ List.tl es}
 
+    | Suspend _, vs ->
+      (* TODO: meaningful timestamp handling *)
+      vs, [e], No_action
+
+  in {t with code = vs', es' @ List.tl es}, action
+
+let rec push_to_inner_stack (c : code) (v : value) : code =
```
rossberg, line 471: Maybe call this `plug`?

interpreter/exec/eval.ml
conrad-watt: I didn't want to bake in an assumption about structural equality being the "right thing" here (what it it becomes possible to have a two memories with different limits but the same backing store?) - is it ok to keep using the `Memory.same` function (renamed as above)?

interpreter/exec/eval.ml
conrad-watt: done

interpreter/exec/eval.ml
conrad-watt: renamed to `plug_value`

conrad-watt:
@rossberg apologies that I let this lie for so long, but PTAL

interpreter/exec/eval.ml
```diff
@@ -57,19 +59,27 @@ and admin_instr' =
   | Label of int32 * instr list * code
   | Frame of int32 * frame * code
 
+type suspend_signal =
+    No_signal
+    (* memory, cell index, target timestamp *)
+  | Wait_signal of memory_inst * Memory.address * float
```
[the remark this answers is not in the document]
conrad-watt: The intention is that in the "long term" this would be fed by `gettimeofday`, which returns a float directly - does it make sense to use `F64.t` here?

interpreter/exec/eval.ml
```diff
@@ -57,19 +59,27 @@ and admin_instr' =
...
+    (* memory, cell index, number of threads to wake *)
+  | Notify_signal of memory_inst * Memory.address * int32
```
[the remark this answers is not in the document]
conrad-watt: done

interpreter/exec/eval.ml
```diff
@@ -57,19 +59,27 @@ and admin_instr' =
   | Label of int32 * instr list * code
   | Frame of int32 * frame * code
 
+type suspend_signal =
```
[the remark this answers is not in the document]
conrad-watt: discarded in current implementation

interpreter/exec/eval.ml
```diff
@@ -57,19 +59,27 @@ and admin_instr' =
...
+  suspend_signal : suspend_signal
 }
 
 type config = thread list
 type thread_id = int
 type status = Running | Result of value list | Trap of exn
 
 let frame inst locals = {inst; locals}
-let thread inst vs es = {frame = frame inst []; code = vs, es; budget = 300}
+let thread inst vs es = {frame = frame inst []; code = vs, es; budget = 300; suspend_signal = No_signal}
```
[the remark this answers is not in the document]
conrad-watt: discarded in current implementation

interpreter/exec/eval.ml
```diff
@@ -57,19 +59,27 @@ and admin_instr' =
...
+  | Wait_signal of memory_inst * Memory.address * float
+    (* memory, cell index, number of threads to wake *)
+  | Notify_signal of memory_inst * Memory.address * int32
+
 type thread =
 {
   frame : frame;
   code : code;
   budget : int;  (* to model stack overflow *)
+  suspend_signal : suspend_signal
```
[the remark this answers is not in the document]
conrad-watt: discarded in current implementation

interpreter/exec/eval.ml
```diff
@@ -57,19 +59,27 @@ and admin_instr' =
   | Label of int32 * instr list * code
   | Frame of int32 * frame * code
 
+type suspend_signal =
+    No_signal
```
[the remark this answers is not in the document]
conrad-watt: discarded in current implementation

conrad-watt:
@rossberg friendly ping

interpreter/exec/eval.ml
```diff
@@ -441,23 +456,70 @@ let rec step_thread (t : thread) : thread =
...
+    let t', act = try step_thread t with Stack_overflow ->
       Exhaustion.error (List.hd (snd t.code)).at "call stack exhausted"
-    in ts1 @ [t'] @ ts2
+    in
+    match act with
+    | NotifyAction (m, addr, count) ->
+      let ts1', count1 = wake ts1 m addr count in
+      let ts2', count2 = wake ts2 m addr (Int32.sub count count1) in
+      ts1' @ [{t' with code = plug_value t'.code (I32 (Int32.add count1 count2))}] @ ts2'
+    | _ -> ts1 @ [t'] @ ts2
```
rossberg, line 522: If WaitAction is not distinguished here, why is it actually needed?

interpreter/exec/eval.ml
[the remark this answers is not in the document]
rossberg: Good question, it should correspond to what we using on paper. Is that a real number?

interpreter/exec/eval.ml
rossberg: Hm, not sure that would make sense. Memories have identity in the semantics, as prescribed by allocating an address for them in the store. The use of `==` (which is pointer equality) is an exact reflection of that.

interpreter/exec/eval.ml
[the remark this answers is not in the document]
conrad-watt: The paper's timestamps would be naturally be rationals. I guess I could use Zarith instead, but this would be an extra dependency.

interpreter/exec/eval.ml
conrad-watt: Changed.

conrad-watt:
@rossberg ptal

interpreter/exec/eval.ml
rossberg: Okay, we may want to think about getting the interpreter closer to the spec somehow wrt the actions it emits, and possibly how it uses them. But let's land this for now.

interpreter/exec/eval.ml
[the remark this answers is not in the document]
rossberg: Yeah, in that case float probably is the closest match.

rossberg (approved):

PR #196, opened 2022-12-12
WIP Threads merge upstream spec

ioannad:
At the moment this is merged up to spec commit f9b461a312426a60f2f81dcb19b39b66b90a5447

The document should build, but not yet the interpreter. 

To try and make sure no threads addition was left out, I created a branch `spec-old` at the latest commit of the upstream spec that was merged previously, (that is I think `7526564b56c30250b66504fe795e9c1e88a938af`), and compared (with `diff` and with `meld`):
- the output of `git diff spec-old threads/main > old-changes.diff`
- with the output of `git diff spec/main threads-update > updated-changes.diff`

So the comparison was done with inspecting `meld old-changes.diff updated-changes.diff` and editing appropriately.


conrad-watt:
The latest changes to the [upstream-rebuild](https://github.com/WebAssembly/threads/tree/upstream-rebuild) branch (tracking issue https://github.com/WebAssembly/threads/issues/198) _should_ mean that this PR is now superceded.

PR #204, opened 2023-07-12
WebAssembly.Memory memory64 constructor parameter proposal

dakenf:
Currently there's no way to create a shared 64bit memory with initial/maximum size>4gb, so an optional parameter can instruct JS engine to check 64 bit limits.

V8 engine is already capable of that and just waits for the spec to be updated, so please let me know if I'm doing it right (update the spec first and then make it working).

`src/wasm/wasm-js.cc`
```
  // TODO(14076): The JS API spec is not updated for memory64 yet; fix this
  // code once it is.
  int64_t initial = 0;
  if (!GetInitialOrMinimumProperty(isolate, &thrower, context, descriptor,
                                   &initial, 0,
                                   i::wasm::kSpecMaxMemory32Pages)) {
    return;
  }
  // The descriptor's 'maximum'.
  int64_t maximum = i::WasmMemoryObject::kNoMaximum;
  if (!GetOptionalIntegerProperty(isolate, &thrower, context, descriptor,
                                  v8_str(isolate, "maximum"), nullptr, &maximum,
                                  initial, i::wasm::kSpecMaxMemory32Pages)) {
    return;
  }
```


rossberg:
I'm afraid this does not belong here. There is a separate [proposal](https://github.com/WebAssembly/memory64) for 64-bit memory, which is currently at phase 3.

conrad-watt:
I think there is an ordering question here. Should the capability to create a shared 64-bit memory live in the memory-64 proposal or in the threads proposal? Do we have a sense of how close memory-64 is to phase 4? I'm currently targetting the October meeting to get threads to phase 4.

rossberg:
Yes, this is one of these intersection-of-2-proposals cases. But before adding any new memory64 extensions in this repo's spec, it would first need to be rebased on the memory64 repo and spec.

As long as we do not know in which order proposals are going to land, I suggest opening a tracking issue (perhaps in both repos) as a reminder that this still needs to be added (here or there) once the order is resolved.

PR #210, opened 2023-09-04
NFC: wip eval

conrad-watt:
major thanks to @ospencer for this


conrad-watt:
With https://github.com/WebAssembly/threads/pull/214, this is now put to rest!

PR #212, opened 2023-09-12
Editorial: try to clarify the JS API

Ms2ger:


document/js-api/index.bs
```diff
@@ -700,12 +710,13 @@ To create a resizable memory buffer we perform the same steps as above, but also
...
-        1. If |memory|.\[[BufferObject]] is an {{ArrayBuffer}},
+    1. If [$IsFixedLengthArrayBuffer$](|buffer|) is true,
+        1. If [$IsSharedArrayBuffer$](|buffer|) is false,
             1. Perform [=!=] [$DetachArrayBuffer$](|buffer|, "WebAssembly.Memory").
-        1. Let |buffer| be the result of [=create a fixed length memory buffer|creating a fixed length memory buffer=] from |memaddr|.
-        1. Set |memory|.\[[BufferObject]] to |buffer|.
+        1. Let |newBuffer| be the result of [=create a fixed length memory buffer|creating a fixed length memory buffer=] from |memaddr|.
+        1. Set |memory|.\[[BufferObject]] to |newBuffer|.
     1. Otherwise,
+        1. Issue: Is this safe for a SharedArrayBuffer?
```
conrad-watt, line 719: Ah, this is a good point. It might not be right to operationally get the "length of |block|" in the SAB case this way.

conrad-watt: I'm currently super jetlagged from Wasmcon but I'll try to think more clearly about the right way to do this SAB case.

conrad-watt (approved):
lgtm, I'll fix the SAB size check issue I commented on above in a follow-up

PR #215, opened 2023-10-04
Preliminary additions for suspensions-consistent-with

ioannad:


conrad-watt:
I ended up pushing something directly that was based on this. Sorry I left this PR dangling!

PR #217, opened 2023-10-18
Fix some typos in `*.wast` tests

alexcrichton:
Most of the tests' `$Check` modules imported from `"mem"` which was failing when I integrated this into Wasmtime. This may very well be a bug in my understanding of the `*.wast` format, however, but currently via `(module $Mem ...)` that defines the name `"Mem"` for modules to use but not `"mem"`. If this is supposed to work as-is I can update Wasmtime instead.

Additionally some tests' comments were checking `L_1` but were actually checking against `local.get 0` so I've updated them to `local.get 1` which fixed some failing tests on Wasmtime.

### DIFF
--- a/test/core/threads/LB.wast
+++ b/test/core/threads/LB.wast
@@ -41,7 +41,7 @@
 (wait $T2)
 
 (module $Check
-  (memory (import "mem" "shared") 1 1 shared)
+  (memory (import "Mem" "shared") 1 1 shared)
 
   (func (export "check") (result i32)
     (local i32 i32)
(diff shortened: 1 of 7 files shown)


rossberg:
I agree that the import from "mem" should fail – no module is registered under that name on the main thread. But likewise "Mem" shouldn't work either – FWIW, the symbolic module name $Mem is unrelated to import names. The correct fix is to add a `(register "mem" $Mem)` to the main thread.

It's a bug that this works in the interpreter as is (I failed to make the registry per-thread correctly). I'll prepare a fix.

It also seems odd that some test succeeds while checking the wrong local. @conrad-watt, any idea?

conrad-watt:
> It's a bug that this works in the interpreter as is (I failed to make the registry per-thread correctly). I'll prepare a fix.

Thanks!

> It also seems odd that some test succeeds while checking the wrong local. @conrad-watt, any idea?

If I'm reading correctly, there are executions where those tests would succeed. I think I simply didn't run those three tests enough times :)

alexcrichton:
Yes of the tests I fixed they were nondeterministically failing locally as well, so can confirm that some executions succeed and others don't

rossberg:
See #219 for the import issue (and others).

alexcrichton:
No worries! No rush on my end

conrad-watt:
Very sorry that I delayed handling this for so long. I've just merged a PR by the wonderful @zilinc (https://github.com/WebAssembly/threads/pull/229) that should fix most of these issues. Please let me know if there are any remaining errors in the tests.

alexcrichton:
Oh nice! I'll close this in favor of that then

Issue #218, opened 2023-10-19
Interpreter can't translate `thread` construct to JS yet


rossberg:
Which means we cannot merge into the main repo yet.

conrad-watt:
There's an approach for doing this in https://github.com/WebAssembly/threads/tree/extract-js-poc but it's super ugly, so I should clean it up (or maybe do a clean slate).

danleh:
V8 also currently doesn't include the tests from this repo due to this missing extraction in the spec interpreter. See https://crbug.com/v8/14376.

rossberg:
@conrad-watt, this proposal still needs some work to move it over the finishing line. ;)

PR #219, opened 2023-10-19
[interpreter] Fix registry to be per-thread

rossberg:
Also:

* Fix module registration in tests to make test pass again
* Fix run.py script to actually run threads tests
* Re-add and extend original `thread.wast` test to check thread scripting features
* Include `either` patterns from main that got lost on upstream-rebuild

Also, I noticed that the JS translation still can't handle the `thread` feature of test scripts. I opened #218.


conrad-watt:
None of the tests are really able to use `either` - it would need to take full result types (i.e. lists of values rather than single values) to support moving the current result logic into the assertion (in general). Are you very attached to it?

rossberg:
I think it was adopted by relaxed SIMD tests, so I don't want to lose its correct integration with the threading implementation.

interpreter/script/js.ml
```diff
@@ -606,8 +628,8 @@ let of_command mods cmd =
     of_assertion' mods act "run" [] None ^ "\n"
   | Assertion ass ->
     of_assertion mods ass ^ "\n"
-  | Thread _ -> failwith "JS translation of Thread is NYI"
-  | Wait _ -> failwith "JS translation of Wait is NYI"
+  | Thread _ -> "" (* TODO: failwith "JS translation of Thread is NYI" *)
```
conrad-watt, line 631: Sorry to have been slow to iterate on this PR. Is this behaviour preferable to having the `failwith` behaviour? My understanding is that this effectively causes garbled JS tests to be silently spat out instead of the translation explicitly failing.

rossberg: The practical problem is that as is, this makes the test runner fail, which also breaks CI. The conceptual problem is that the current assumption is that every core test can be converted to JS. If we want to exempt some tests, we'll need additional infrastructure for that. But what would that imply for the role of the test suite?

Is there truly no way to transform the thread construct to JS faithfully?

conrad-watt: There is - it would have to be a follow-up PR though. I'm just trying to clarify if generating garbled tests would be a preferable behaviour in the meantime.

conrad-watt: Could we at least turn the thread construct into some kind of dynamic "assert false" instead?

rossberg: Ah, I see. The idea of opening an issue for it was to do it separately. Adding a "correct" temporary work-around probably is more work than is worthwhile.

rossberg: (The thread construct is in a completely different category than the arguments of assertions, so we cannot simply wrap it into one.)

Issue #216, opened 2023-10-12
Move the `upstream-rebuild` branch to `main`


alexcrichton:
This was discussed a bit during the CG meeting yesterday but I wanted to open an issue for this as well. Downstream consumers such as the [testsuite repository](https://github.com/WebAssembly/testsuite) pick up changes to tests from the `main` branch, and additionally when visiting the repository it's useful to see the current state on the landing page, currently the `main` branch as well.

If I might offer a concrete suggestion, it might be useful to rename `main` to `old-main` (please bikeshed) and then rename the `upstream-rebuild` branch to `main` to avoid losing histories.

Issue #184, opened 2021-12-01
Spec draft text format for memory type inconsistent with tests


takikawa:
The draft spec at https://webassembly.github.io/threads/core/text/types.html#text-memtype shows the `shared` modifier for memory as starting a new nesting level:

![image](https://user-images.githubusercontent.com/64192/144329756-811f1a05-59d6-4385-a966-0c9970727cd3.png)

However, this doesn't match the examples or tests where `shared` is just a keyword that can come after the limits.

e.g., `(module (import "spectest" "shared_memory" (memory 1 2 shared)))` from https://github.com/WebAssembly/threads/blob/main/test/core/imports.wast#L499

rossberg:
Thanks, I fixed the grammar.

rossberg:
Hm, I had to search for it. Seems like I still [committed](https://github.com/WebAssembly/threads/commit/b438715e7edfb026dc8945f47e5fc18edf26ecbf) it to master at the time, but somehow it never reached main. No idea why. I just merged it over manually.

PR #221, opened 2023-12-20
Fix wat in Overview.md

tlively:
The definition of the shared memory did not follow the proposed text format.

### DIFF
--- a/proposals/threads/Overview.md
+++ b/proposals/threads/Overview.md
@@ -16,7 +16,7 @@ mutex is unlocked. If its value is 1, the mutex is locked.
 ```wasm
 (module
   ;; Import 1 page (64Kib) of shared memory.
-  (import "env" "memory" (memory 1 1 shared))
+  (import "env" "memory" (memory (shared 1 1)))
 
   ;; Try to lock a mutex at the given address.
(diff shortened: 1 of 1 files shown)


tlively:
It doesn't look like the GitHub Pages version of the threads spec has been updated, btw, so I had to go to the source to find the real proposed text format.

conrad-watt:
@tlively what is the source you were looking at to find that proposed syntax? As far as I can tell the text format specification currently just doesn't include any specification for the shared annotation anywhere, including in the raw source.

I also think that `memory 1 1 shared` is the syntax we agreed on, and IIUC I should update the text format specification to reflect this.

rossberg:
@conrad-watt, it [does here](https://github.com/WebAssembly/threads/blob/main/document/core/text/types.rst#memory-types).

conrad-watt:
Ah, I was looking at the upstream-rebuild branch. Probably another reason to move ahead with https://github.com/WebAssembly/threads/issues/216...

tlively:
Oh yeah I forgot about all the different branches. Please put the real stuff on `main`!

tlively:
@rossberg, apparently you already fixed the spec according to #184. Do you know where that fix went?

PR #227, opened 2024-08-07
Fix typo in Compare Exchange table

jemc:
Prior to this commit there was a copy/paste typo in the Compare Exchange table wherein a 16-bit operation is said in one column to be 8-bit (pasted from the line above).

This commit fixes the typo so that it reflects the correct number of bits.


conrad-watt (approved):
thanks!

PR #235, opened 2025-10-06
Remove duplicate (broken) definition of `assert_ArrayBuffer`

backes:
This was removed from the upstream test in
https://github.com/WebAssembly/spec/commit/cc2d59bd56e5342e3c1834a7699915f8b67fc29c.

The `assert_ArrayBuffer` helper is now defined in the `assertions.js` file. The version there does not have the problem with `self`.

This problem was uncovered when importing the threads JS tests into V8, see https://crbug.com/449581914.

### DIFF
--- a/test/js-api/memory/grow.any.js
+++ b/test/js-api/memory/grow.any.js
@@ -1,30 +1,6 @@
 // META: global=window,dedicatedworker,jsshell
 // META: script=/wasm/jsapi/memory/assertions.js
 
-function assert_ArrayBuffer(actual, expected, message) {
-  // https://github.com/WebAssembly/spec/issues/840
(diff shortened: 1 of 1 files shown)


backes:
Alternatively to this targeted fix we could "just" rebase this repository, I guess...

conrad-watt:
@backes is this fix already present in the https://github.com/WebAssembly/threads/tree/upstream-rebuild of this repository? That branch contains a rebase. If so, I should probably rename that branch to "main", which I've intended to do for a while but never got around to.

conrad-watt:
I think most of the testing was done in the browser rather than through a headless shell, so there may be some quirks to iron out. I'll keep most of the follow-ups in the other issue.

conrad-watt:
closing now we've rebased, but we can discuss JS support issues separately

PR #236, opened 2025-10-06
Fix proxy used in memory constructor test

backes:
Another targeted fix to make this test pass in V8. This was fixed upstream already, so alternatively we could rebase the whole repository.


conrad-watt:
Same comment as https://github.com/WebAssembly/threads/pull/235#issuecomment-3372190914

PR #242, opened 2026-01-29
Make non-natural alignment a validation error in spec interpreter / test

stevenfontanella:
Followup to #243. Prior to #243, a non-natural alignment for an atomic instruction was *not* a parse, validation or runtime failure in the spec interpreter / tests, although the spec specified that it would be a parse error (and would be a validation error after #243). Change the reference interpreter and spec test to make a non-natural alignment into a validation error. 8-bit instructions aren't tested for the validation failure since there's no way to have a non-natural alignment for those cases (only a greater one which results in a different validation failure).

Also removed some extra spaces in cmpxchg tests.

Tested with `make test` locally.


test/core/threads/atomic.wast
```diff
@@ -437,6 +437,131 @@
 (assert_trap (invoke "i64.atomic.rmw16.cmpxchg_u" (i32.const 1) (i64.const 0) (i64.const 0)) "unaligned atomic")
 (assert_trap (invoke "i64.atomic.rmw32.cmpxchg_u" (i32.const 1) (i64.const 0) (i64.const 0)) "unaligned atomic")
 
+;; Alignment hint is lower than the natural alignment. Execution will trap even if the address being read is aligned.
```
tlively: Since this is a validation error rather than an execution error, we need to put each instruction in its own module with its own separate `assert_invalid`. Otherwise correctly implementing the validation for just one instruction would mask missing the validation for all the other instructions.

tlively (approved):
LGTM. Let's see what @conrad-watt and @rossberg think.

rossberg (commented):
Looks good, I only have stylistic nits regarding the validator change.

interpreter/valid/valid.ml
```diff
@@ -189,17 +189,22 @@ let check_vec_binop binop at =
...
       packed_size sz
-  in
-  require (1 lsl memop.align <= size) at
-    "alignment must not be larger than natural"
+  in match mode with
+    | NonAtomic ->
+        require (1 lsl memop.align <= size) at
+          "alignment must not be larger than natural";
+    | Atomic ->
+        require (1 lsl memop.align == size) at "atomic memory instruction's alignment must equal the instruction's natural alignment"
```
rossberg:
```suggestion
  in
  match mode with
  | NonAtomic ->
    require (1 lsl memop.align <= size) at
      "alignment must not be larger than natural";
  | Atomic ->
    require (1 lsl memop.align = size) at
      "atomic alignment must be natural"
```

stevenfontanella:
Opened #244 to see if the CI can be fixed easily but I don't have permission to run it. Can you check if that helps?

rossberg (approved):

PR #248, opened 2026-03-12
[threads][test][js-api] Add tests for growing shared memory

backes:
Add a few first (!) js-api test to this repo. In particular:
- Growing a non-shared memory always detaches the old buffer and allocates a fresh one (as before).
- Growing a shared memory only refreshes the buffer if the delta in pages is > 0.
- Growing a shared memory has no effect on unrelated shared memories.

Note that V8 didn't implement this correctly until https://crrev.com/c/7660970, see https://crbug.com/492128992.

### DIFF
--- a/test/js-api/memory/grow.any.js
+++ b/test/js-api/memory/grow.any.js
@@ -187,3 +187,40 @@ test(() => {
   assert_equals(newArray[0], 1, "new first element");
 
 }, "Growing shared memory does not detach old buffer");
+
+test(() => {
+  const argument = { "initial": 1 };
+  const memory = new WebAssembly.Memory(argument);
+  const oldMemory = memory.buffer;
+
+  const result = memory.grow(0);
+  assert_equals(result, 1);
+
+  const newMemory = memory.buffer;
+  assert_not_equals(oldMemory, newMemory, "Buffer identity must NOT be preserved on grow(0)");
+  assert_ArrayBuffer(oldMemory, { "detached": true }, "Buffer must be detached on grow(0)");
+}, "Growing non-shared memory by 0 does change the buffer");
+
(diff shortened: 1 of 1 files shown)


tlively:
I don't see where the spec says to check for delta > 0 before refreshing the buffer.

https://webassembly.github.io/threads/js-api/index.html#grow-the-memory-buffer

backes:
Oh, I was basing this on this algorithm in the overview: https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#webassemblymemoryprototypebuffer

But now I see that this is totally different to the growing algorithm and the `buffer` accessor as spec'ed in the JS-API. So I guess the spec wins?

conrad-watt:
Urgh, sorry about this divergence. Experimentally, what do engines do right now?

backes:
It looks like all browsers implement the algorithm in the overview, not the spec :D 

I attached an HTML page for testing and a little python script for serving the HTML file with the right COOP and COEP headers (vibe coded):
[test-shared-wasm-memory-growth.html](https://github.com/user-attachments/files/27005277/test-shared-wasm-memory-growth.html)
[server.py](https://github.com/user-attachments/files/27005278/server.py)

Results:
Chrome 149, Firefox 150, Sarafi 26.4 (and tech preview) all agree on the behavior as described in the overview. Yay!

(Note that Chrome 147 behaved differently and refreshed *all* ArrayBuffers for *all* shared Wasm memories in the isolate; this was fixed by https://crrev.com/c/7660970.)

backes:
I opened a PR to adapt the JS-API spec to match what's described in the overview: https://github.com/WebAssembly/threads/pull/255

PR #250, opened 2026-03-17
[threads][test][js-api] Add tests for wait/notify

backes:
Add JS API tests for Atomics.wait and Atomics.notify on WebAssembly.Memory.
- On shared memory: Atomics.notify should return the number of notified waiters, and Atomics.wait should return the appropriate wait status.
- On unshared memory: Atomics.notify should return 0, while Atomics.wait should trap (throwing a WebAssembly.RuntimeError).


conrad-watt:
LTGM!

backes:
Thanks for merging, Conrad!
Note that there are a few more: #248, #249, and #251.

PR #251, opened 2026-03-17
[threads][test][js-api] Add worker-based tests for Wait/Notify

backes:
Extend JS API tests for Atomics.wait and Atomics.notify with worker-based
scenarios. These tests verify that workers can be correctly woken up by
notify calls on shared WebAssembly.Memory.

Specifically:
- Add worker-path-helper.js to handle worker script path resolution in
  different environments (WPT vs JS shells).
- Add wait-notify-worker.js to execute the wait operation in a background
  thread.
- Add test cases to wait-notify-shared.any.js covering single waiter,
  sequential multiple waiters, and simultaneous multiple waiters.

### DIFF
--- a/test/js-api/memory/wait-notify-shared.any.js
+++ b/test/js-api/memory/wait-notify-shared.any.js
@@ -1,5 +1,6 @@
 // META: global=window,dedicatedworker,jsshell
 // META: script=/wasm/jsapi/wasm-module-builder.js
+// META: script=/wasm/jsapi/memory/worker-path-helper.js
 
 function createModule() {
   const builder = new WasmModuleBuilder();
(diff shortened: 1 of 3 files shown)


backes:
This PR is based on https://github.com/WebAssembly/threads/pull/250 and extends the test. I don't know how to create such stacked PRs on Github. The first commit is the same as in https://github.com/WebAssembly/threads/pull/250, the second one adds new worker-based tests.

I didn't find precedence for such worker tests but tried my best to make them work in all environments. This might need further adjustments if some shells or browsers run into issues, potentially after syncing this to WPT.

After https://github.com/WebAssembly/threads/pull/250 got merged we can rebase this PR and it will only contain the added lines and files.

backes:
Rebased, now that #250 has been merged.

